PR #7, opened 2021-04-01
Display notice when users record or transcribe meeting

cheukchuen:
## Purpose

Display notification banner when users record and/or transcribe meeting

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -261,6 +261,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
     private func onJoinCall() {
         NotificationCenter.default.addObserver(self, selector: #selector(onRemoteParticipantsUpdated(_:)), name: .remoteParticipantsUpdated, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(recordingActiveChangeUpdated(_:)), name: .onRecordingActiveChangeUpdated, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(transcriptionActiveChangeUpdated(_:)), name: .onTranscriptionActiveChangeUpdated, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(onCallStateUpdated(_:)), name: .onCallStateUpdated, object: nil)
         onCallStateUpdated()
         meetingInfoViewUpdate()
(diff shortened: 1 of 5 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -521,10 +522,46 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+            case (true, true):
+                return meetingRecordingAndTranscriptionActiveText
+            case (false, true):
+                return meetingRecordingStopTranscriptionActiveText
+            default:
+                return meetingRecordingStopText
+            }
+        }()
+
+        messageBannerView.showBannerMessage(notificationText)
```
waynemo, line 543: We can create a private func and reuse this logic for both.

waynemo (changes requested):

AzureCalling/Controllers/CallViewController.swift
AmyL219: Not sure if one logic function can be applied to the events.
Since the text is different based on triggering event. For example, if isRecordingActive = true, isTranscriptionActive = false. If it is in the Recording listener, we would display "Recording start", if it's trigger from Transcription change, it would display "Transcription stops but recording is continuing"

waynemo: We always update the message based on the values of both `isRecordingActive` and `isTranscriptionActive`, and at this point those values are already updated in `CallingContext`. So we should be able to reuse the same logic.

cheukchuen: Updated with a helper function. Also changed the base branch to target `feature/teams_interop`

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -520,14 +521,44 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         meetingInfoViewUpdate()
     }
 
-    @objc func recordingActiveChangeUpdated(_ notification: Notification) {
-        guard let isRecordingActive = callingContext?.isRecordingActive else {
+    private func toggleMeetingCaptureNotification(recording: NSAttributedString, transcribing: NSAttributedString, bothActive: NSAttributedString, stopped: NSAttributedString) {
+        guard let isRecordingActive = callingContext?.isRecordingActive,
```
AmyL219, line 525: I think we could combine the word logic into the toggleNotification function
You could have a look on Android side implement method (use an enum as input) or implement a simple `state machine`: https://albertodebortoli.com/2018/12/16/the-easiest-state-machine-in-swift/

cheukchuen: Could you please clarify the Android side implementation method using enums? I'm seeing basic implementation using [if-else logic](https://github.com/Azure-Samples/communication-services-android-calling-hero/blob/2d529525085f759fa02f16efa7431f742d5a0170/AzureCalling/app/src/main/java/com/azure/samples/communication/calling/activities/CallActivity.java#L378-L404).

Also wondering if we are to use state machines, could we just utilize the [Stateful](https://github.com/albertodebortoli/Stateful) package by including `pod 'Stateful'` in the podfile?

cheukchuen: I've changed the `toggleMeetingCaptureNotification` method to accept an enum. That way if there are more types added in the future, the switch statement grows instead of the parameter list of a function.

Personally I would prefer an actual state machine implementation, but for now that might be an overkill.

AmyL219 (dismissed):
Looks good to me

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -520,12 +526,40 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+            case (.recording, false, false):
+                return meetingRecordingStopText
+            case (.transcription, true, false):
+                return meetingRecordingActiveTranscriptionStopText
+            case (.transcription, false, true):
+                return meetingTranscriptionActiveText
+            case (.transcription, false, false):
+                return meetingTranscriptionStopText
+            }
+        }()
```
waynemo, line 552: Creating this closure every time just to return the text does not feel right. I think we can just assign the `notificationText` variable to simplify it.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -35,7 +40,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
     @IBOutlet weak var toggleVideoButton: UIButton!
     @IBOutlet weak var toggleMuteButton: UIButton!
     @IBOutlet weak var infoHeaderView: InfoHeaderView!
-    @IBOutlet weak var messageBannerView: MessageBannerStackView!
+    @IBOutlet weak var messageBannerStackView: MessageBannerStackView!
```
waynemo, line 43: Since we are using the name NoticeBannerType, I think we can rename this to `noticeBannerStackView` and `NoticeBannerStackView` to be consistent.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -520,12 +526,40 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         meetingInfoViewUpdate()
     }
 
-    @objc func recordingActiveChangeUpdated(_ notification: Notification) {
-        guard let isRecordingActive = callingContext?.isRecordingActive else {
+    private func toggleMeetingCaptureNotification(mode: NoticeBannerType) {
```
waynemo, line 529: To align with the naming, we can call this `showNoticeBanner`.

waynemo (changes requested):

waynemo (approved):
Looks good to me.

Let's also make sure to squash merge to feature/teams_interop and clean up some of the intermediate messages.

AmyL219 (approved):

PR #8, opened 2021-04-01
Added pausing and resuming the video during a call when app is put into the background

Gemakk:
## Purpose

*  During a call, if the video is on:
  * Pause the local video to remote participants if the app is brought out of focus (Into the background, or covered by the Notification Center)
  * Resume the local video to remote participants once focus is returned

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -544,6 +546,16 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         }
     }
 
+    @objc func appOutOfFocus(_ notification: Notification) {
+        print("appOutOfFocus")
+        callingContext.pauseVideo()
+    }
+
```
AmyL219, line 553: It's better to rename these two functions similar to `appResignActive` and `appAssignActive`. 
In iOS, resign/assign Active means app is no longer in foreground and the opposite

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -183,6 +196,18 @@ class CallingContext: NSObject {
...
+            stopVideo { [weak self] _ in
+                guard let self = self else {
+                    return
+                }
+                self.isVideoOnHold = true
+                print("Local video paused successfully")
+            }
+        }
+    }
+
```
AmyL219, line 210: I think it would be better to rename those functions to more specific name
-startLocalVideoStream
-stopLocalVideoStream
-pauseLocalVideoStream
-resumeLocalVideoStream

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -170,6 +171,18 @@ class CallingContext: NSObject {
         }
     }
 
+    func resumeVideo() {
+        if isVideoOnHold {
```
AmyL219, line 175: We could add async completionHandler. Could take a reference from `endCall`
`resumeVideo(completionHandler: @escaping (Result<Void, Error>) -> Void)`

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -183,6 +196,18 @@ class CallingContext: NSObject {
         }
     }
```
AmyL219, line 198: We could add async completionHandler.

AmyL219 (changes requested):
Just a few comments

AmyL219 (approved):

cheukchuen (approved):
Functionally tested, seems to be working as expected

PR #15, opened 2021-04-07
Include mute indicator on display name when toggled

cheukchuen:
## Purpose

Include mute indicator on display name when toggled by either local or remote participants

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
- toggle mute / unmute in waiting room when start new call
- toggle mute / unmute in 1:1 group call then add / remove participants to see if your own mute status is correct
- toggle mute / unmute in 3+ participants
- for remote participants in active display, toggle their mute / unmute
- for remote participants not in active display, toggle their mute / unmute, then bring them into active display to see if their mute status is correct
- see if all participants have correct mute status if join another call
- any other scenarios not stated here


AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -334,6 +338,13 @@ class CallingContext: NSObject {
     private func setupRemoteParticipantsEventsAdapter() {
         participantsEventsAdapter = ParticipantsEventsAdapter()
 
+        participantsEventsAdapter?.onIsMutedChanged = { [weak self] _ in
+            guard let self = self else {
+                return
+            }
+            self.notifyRemoteParticipantsUpdated()
```
AmyL219, line 345: 1. Can we do a filter here to see if the participant with muting state change in displayed in the calling view grid.
 If he is not in the grid, we don't need to notify the views updates.  
2. Would it be kind of waste of resource that every time a small muting icon change on one participant view triggers all participant view updating? Is it possible to implement business logic to find the participant view need updates and only update that specific view in the Calling View controllers?

AzureCalling/Views/ParticipantView.swift
```diff
@@ -37,8 +37,21 @@ class ParticipantView: UIView {
 
     // MARK: Operation methods
 
-    func updateDisplayName(displayName: String) {
-        participantLabel.text = displayName
+    func updateDisplayName(displayName: String, isMuted: Bool) {
```
AmyL219, line 40: 1. Would it be more reasonable that we have a separate api to update muting state icon based on the single responsibility?
2. Instead of appending an NSTextAttachment, we could add an icon in the participantNameLabel, and one separated api interface to toggle the icons change. Since our project is more storyboard based and we are trying to make less UI business code in the code

AmyL219 (commented):
A few thoughts. Let me know if you have any question

AzureCalling/Views/ParticipantView.swift
cheukchuen: Agreed on the separate api. Made the change to utilize storyboard for UI logic

AzureCalling/External/Calling/CallingContext.swift
cheukchuen: Added another event to target only active participant that is on screen if triggered and update the participant's mute indicator

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -235,8 +235,22 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
     @IBAction func onToggleMute(_ sender: UIButton) {
+        let updateLocalDisplayName: (Result<Void, Error>) -> Void = { [weak self] toggleMuteResult in
+            guard let self = self else {
+                return
+            }
+            DispatchQueue.main.async {
+                switch toggleMuteResult {
+                case .success:
+                    self.localParticipantView.updateDisplayName(displayName: self.callingContext.displayName + " (Me)", isMuted: self.callingContext.isMuted)
+                case .failure:
```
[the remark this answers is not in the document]
cheukchuen: No longer needed to abstract away into a reusable function since we've separated the API interface

AzureCalling.xcodeproj/project.pbxproj
```diff
@@ -87,6 +88,7 @@
 		6D911F99B1E08BC60AF9272D /* Pods-AzureCalling.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-AzureCalling.debug.xcconfig"; path = "Target Support Files/Pods-AzureCalling/Pods-AzureCalling.debug.xcconfig"; sourceTree = "<group>"; };
 		6E8B22F0F730B9A7FF0B41EE /* Pods-AzureCalling.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-AzureCalling.release.xcconfig"; path = "Target Support Files/Pods-AzureCalling/Pods-AzureCalling.release.xcconfig"; sourceTree = "<group>"; };
 		83FB20DFF3EE83EF2044575D /* Pods-ACSCall.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-ACSCall.release.xcconfig"; path = "Target Support Files/Pods-ACSCall/Pods-ACSCall.release.xcconfig"; sourceTree = "<group>"; };
+		987C3DC52624DE6200E296D0 /* ParticipantLabelView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ParticipantLabelView.swift; sourceTree = "<group>"; };
```
AmyL219, line 91: Do we want to keep ParticipantLabel? Would it be overlapped with ParticipantLabelView? Since we can use ParticipantLabelView to setup the padding/background of the label.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -236,7 +236,22 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     @IBAction func onToggleMute(_ sender: UIButton) {
         sender.isSelected = !sender.isSelected
-        sender.isSelected ? callingContext.mute { _ in} : callingContext.unmute { _ in}
+        sender.isSelected ? callingContext.mute(completionHandler: updateMuteIndicator) : callingContext.unmute(completionHandler: updateMuteIndicator)
+    }
+
+    private func updateMuteIndicator(_ result: Result<Void, Error>) {
+        DispatchQueue.main.async { [weak self] in
```
AmyL219, line 243: Perhaps renaming into `updateLocalMuteIndicator(isMuted: Bool)`. the parameter `Result` doesn't seem necessary
You can rewrite the `onToggleMute` function to make it more readable

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -510,6 +529,18 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         meetingInfoViewUpdate()
     }
 
+    @objc func onRemoteParticipantIsMutedChanged(_ notification: Notification) {
```
AmyL219, line 532: My concern here is if pass RemoteParticipant through notification is sensible. 
Usually we don't pass an object in the notification, instead we could do `delegate`, `closure`.
My original thought is 1. check if the remoteParticipant is in the list 2.0 If no, no notification post. 2.1 If yes, post notification 3. All participant view will get updates. 
Please chat with Pavel to sync the thought here.

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -16,6 +16,7 @@ class CallingContext: NSObject {
     // MARK: Properties
     private (set) var joinId: String!
     private (set) var displayName: String!
+    private (set) var isMuted: Bool = false
```
AmyL219, line 19: I dont see we need this property right now. Remember me if I am wrong. 
We can use the call.isMute in CallingSDK 12.

AzureCalling/Views/ParticipantLabel.swift
```diff
@@ -9,13 +9,6 @@ import UIKit
 class ParticipantLabel: UILabel {
```
AmyL219, line 9: Let's try to move those UI configuration into Storybroad configuration in ParticipantLabelView

AzureCalling/Views/ParticipantLabelView.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@IBDesignable
+class ParticipantLabelView: UIStackView {
+    var insets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 0)
+
```
AmyL219, line 11: What's the insets for?

AzureCalling/Views/ParticipantLabelView.swift
```diff
@@ -0,0 +1,18 @@
...
+    @IBInspectable var cornerRadius: CGFloat = 0 {
+        didSet {
+            layer.cornerRadius = cornerRadius
+            layer.masksToBounds = cornerRadius > 0
```
AmyL219, line 15: What's the purpose of `layer.masksToBounds = cornerRadius > 0`?

AzureCalling/Views/ParticipantView.swift
```diff
@@ -22,6 +22,7 @@ class ParticipantView: UIView {
     @IBOutlet weak var placeholderImage: UIImageView!
     @IBOutlet weak var videoViewContainer: UIView!
     @IBOutlet weak var participantLabel: ParticipantLabel!
+    @IBOutlet weak var participantMuteIndicator: UIView!
```
AmyL219, line 25: Let's replace participantLabel with participantLabelView. And remove participantMuteIndicator.
The `participantMuteIndicator` is a subview inside `ParticipantLabelView`. We should link it in `ParticipantLabelView` class. And open a `api` in `ParticipantLabelView` to `updateMuteIndicator`

AzureCalling/Views/ParticipantView.swift
```diff
@@ -42,6 +43,10 @@ class ParticipantView: UIView {
         participantLabel.isHidden = displayName.isEmpty
     }
 
+    func updateMuteIndicator(isMuted: Bool) {
+        participantMuteIndicator.isHidden = !isMuted
```
AmyL219, line 47: Let's open an api in ParticipantLabelView to updateMuteIndicator. 
And call `participantLabelView.updateMutesIndicator`  here

AmyL219 (commented):
It's much better. Let me know if you have any question. 
A general question, when user come into a new group call, does `OnIsMutedChanged` for  every remoteParticipant get called? AKA. how do we initial the mute state for every new added `remoteParticipants`?

AzureCalling/Controllers/CallViewController.swift
AmyL219: And in 1:1 group call, do we need to update local mute indicator?

AzureCalling/Views/ParticipantView.swift
cheukchuen: Good suggestion, this will have separation of concerns

AzureCalling/Views/ParticipantLabelView.swift
cheukchuen: Removed. Accidentally copied this over from ParticipantLabel

AzureCalling/Views/ParticipantLabel.swift
cheukchuen: Moved the padding setup to ParticipantLabelView in Storyboard using `Layout Margins`

AzureCalling.xcodeproj/project.pbxproj
cheukchuen: Most visual aspects of the container is being done in ParticipantLabelView through Storyboard

AzureCalling/Views/ParticipantLabelView.swift
cheukchuen: This was copied over from the previous ParticipantLabel. The purpose of `masksToBounds` is to clip the layer to its boundaries, though it does not make any difference in our case. Removed it to simplify stuff

AzureCalling/External/Calling/CallingContext.swift
cheukchuen: I'll make the appropriate changes after the main branch has been updated to use CallingSDK 12

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Will work on this after the main branch has been updated to use CallingSDK 12 since it relates to `call.isMuted`

Good point on the 1:1 group call though, seems that we can't see our own label. There's no need to update the indicator if it's 1:1

AzureCalling/Controllers/CallViewController.swift
cheukchuen: I've tried both targeted updates and update all approach. The former seems instantaneous whereas the latter can have some delay in the updates. Somehow we do need to pass data for the next task when we want to identify active speaker.

So I've looked through stackoverflow and some blogs. It's fine to use notification to pass data, especially useful when it's broadcasting from one class to multiple classes, although we need to pass in the correct arguments to [post(name:object:userInfo:)](https://developer.apple.com/documentation/foundation/notificationcenter/1410608-post) for it to be semantically correct. `object` applies to the sender of the notification, whereas `userInfo` is used to pass additional data which we can utilize.

cheukchuen:
> A general question, when user come into a new group call, does `OnIsMutedChanged` for every remoteParticipant get called? AKA. how do we initial the mute state for every new added `remoteParticipants`?

For every newly added `remoteParticipants`, `onRemoteParticipantsUpdated` is triggered to update all the participants view. `onIsMutedChanged` is specifically for existing participants toggling the mute button during the call.

AzureCalling/Controllers/CallViewController.swift
waynemo: I think we should update all for now, and optimize targeting specific participant views separately. Unless there is something causing extra time for each participant view update, there should not be a noticeable delay since we are updating 4 max.

AmyL219:
> > A general question, when user come into a new group call, does `OnIsMutedChanged` for every remoteParticipant get called? AKA. how do we initial the mute state for every new added `remoteParticipants`?
> 
> For every newly added `remoteParticipants`, `onRemoteParticipantsUpdated` is triggered to update all the participants view. `onIsMutedChanged` is specifically for existing participants toggling the mute button during the call.

If user A join a meeting where user B already joined and user B is in the participant grid. And if user B muted his voice during all the meeting, can user A see user B is muted? How is this scenario handled?

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Removed `updateLocalMuteIndicator(isMuted: Bool)` altogether after refactoring to use `call.isMuted`.

About 1:1 group call, tried to not update local mute indicator, but doesn't work as expected since any changes to the participant views would result in outdated local participant label. We still need to update it in 1:1 group call.

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Changed to update just the mute indicator for all remote participants if there're any mute toggling events.

cheukchuen:
> > > A general question, when user come into a new group call, does `OnIsMutedChanged` for every remoteParticipant get called? AKA. how do we initial the mute state for every new added `remoteParticipants`?
> > 
> > 
> > For every newly added `remoteParticipants`, `onRemoteParticipantsUpdated` is triggered to update all the participants view. `onIsMutedChanged` is specifically for existing participants toggling the mute button during the call.
> 
> If user A join a meeting where user B already joined and user B is in the participant grid. And if user B muted his voice during all the meeting, can user A see user B is muted? How is this scenario handled?

1. First, this function gets called `call(_ call: Call, didUpdateRemoteParticipant args: ParticipantsUpdatedEventArgs)` to clear and update the participants view.
2. Then the event `.remoteParticipantIsMutedChanged` is being triggered for the number of existing remote participants in the meeting room to update their mute indicators.
3. After that, this function gets called `call(_ call: Call, didChangeState args: PropertyChangedEventArgs)` to update the participants view.

AzureCalling/Controllers/CallViewController.swift
waynemo: What I meant was we should just be able to call `queueParticipantViewsUpdate()`.

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -223,27 +223,27 @@ class CallingContext: NSObject {
...
+    func unmute(completionHandler: @escaping (Result<Call?, Error>) -> Void) {
         self.call?.unmute(completionHandler: { (error) in
             if error != nil {
                 print("ERROR: It was not possible to unmute. \(error!)")
                 completionHandler(.failure(error!))
                 return
             }
             print("Unmute successful")
-            completionHandler(.success(()))
+            completionHandler(.success(self.call))
```
waynemo, line 246: We should not need to return the `Call` object.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -236,7 +236,17 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     @IBAction func onToggleMute(_ sender: UIButton) {
         sender.isSelected = !sender.isSelected
-        sender.isSelected ? callingContext.mute { _ in} : callingContext.unmute { _ in}
+        (sender.isSelected ? callingContext.mute : callingContext.unmute) { [weak self] result in
+            guard let self = self else {
+                return
+            }
+            if case .success(let call) = result,
+               let isMuted = call?.isMuted {
```
waynemo, line 244: We shouldn't need the `Call` object here. We should be able to update the mute indicator based on success/error.

waynemo (dismissed):

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Made use of `sender.isSelected` instead

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Changed to use `queueParticipantViewsUpdate()`, though there's a 1-2 seconds delay from my experience

AmyL219:
A general test on the muting function **without** test on 
`for remote participants not in active display, toggle their mute / unmute, then bring them into active display to see if their mute status is correct`

AmyL219 (approved):

palatter (approved):

PR #16, opened 2021-04-09
Move isCameraPreferredOn to update in closure

cheukchuen:
## Purpose

* Only update isCameraPreferredOn when start/stop video succeeds

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/AzureCalling/External/Calling/CallingContext.swift
+++ b/AzureCalling/External/Calling/CallingContext.swift
@@ -135,7 +135,6 @@ class CallingContext: NSObject {
     }
 
     func startLocalVideoStream(completionHandler: @escaping (LocalVideoStream?) -> Void) {
-        isCameraPreferredOn = true
         withLocalVideoStream { [weak self] localVideoStream in
             guard let self = self else {
                 return
@@ -149,8 +148,10 @@ class CallingContext: NSObject {
             self.call?.startVideo(stream: localVideoStream) { (error) in
                 if error != nil {
                     print("ERROR: Local video failed to start. \(error!)")
+                    completionHandler(nil)
                     return
                 }
+                self.isCameraPreferredOn = true
                 print("Local video started successfully")
             }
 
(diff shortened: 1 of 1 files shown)


Gemakk (approved):

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -149,8 +148,10 @@ class CallingContext: NSObject {
             self.call?.startVideo(stream: localVideoStream) { (error) in
                 if error != nil {
                     print("ERROR: Local video failed to start. \(error!)")
+                    completionHandler(nil)
```
AmyL219, line 151: Do we need this? We pass `completionHandler(self.localVideoStream)` before the CallingSDK callback closure is being called.  
This error apply to `calling SDK`'s failure in transporting local video stream, but locally we have the `localVideoStream`. If the callingSDK failed to pass the local video stream to SDK, do we want to display the self local video to the user.  
Same question apply to the `isCameraPreferredOn` movement. Do we want to trigger the flag only when the `callingSDK` successfully pass our local video?
@waynemo, what's your thought

waynemo: I think the intent of the flag is to indicate the preferred on/off state of the camera so that the UI can show the preferred selection. For example:

- If isCameraPreferredOn=false, the toggle button should show the "camera off" icon, and the local participant view would display the avatar image instead of the RendererView.
- If isCameraPreferredOn=true, the toggle button should show the "camera on" icon, and the local participant view would display the RendererView instead of the avatar image. If the localVideoStream is available, I think we should still be showing the local video since it is not the issue. 

The main concern would be how to indicate that the local video is not actually being sent. Let's discuss more offline.

AmyL219:
As per discussion, no handling would apply in sample app for now.

PR #17, opened 2021-04-12
Update new colours for buttons, textFields

AmyL219:
## Purpose

* Update colours of primary, gray 300 to support the UI requirement in dark mode

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


Gemakk (dismissed):

cheukchuen:
Just to confirm whether the lighter shade of blue is the intended color

cheukchuen (commented):
Just wanted to confirm whether the lighter shade of blue is the intended color

AmyL219:
> Just wanted to confirm whether the lighter shade of blue is the intended color

I updated the primary color to the hex value `#0078D4` which should be correct value from Figma

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -169,6 +167,25 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
...
+        startCallButton.setTitle(startButtonTitle, for: .normal)
+        if let icon = UIImage(named: "ic_fluent_meet_now_24_regular") {
+            let buttonIcon = icon.withRenderingMode(.alwaysTemplate)
+            startCallButton.tintColor = UIColor.systemBackground
+            startCallButton.setImage(buttonIcon, for: .normal)
+        }
+    }
+
+    private func setupNameTextField() {
+        let placeJHolder = "John Smith"
```
cheukchuen, line 182: Typo

cheukchuen (commented):
Found a typo.

Also is it preferable if we use the term AccentColor instead of PrimaryColor? AccentColor is in line with the term used in SwiftUI's [accentColor(_:)](https://developer.apple.com/documentation/swiftui/color/accentcolor(_:)). This should also eliminate the warning issued by Xcode regarding `Accent color 'AccentColor' is not present in any asset catalogs.` Otherwise if PrimaryColor is preferred, then we should update `[AzureCalling|Pods].xcodeproj`to point the necessary accent fields to PrimaryColor.

AzureCalling/Controllers/LobbyViewController.swift
AmyL219: Good catch. Fixed

AmyL219:
> Found a typo.
> 
> Also is it preferable if we use the term AccentColor instead of PrimaryColor? AccentColor is in line with the term used in SwiftUI's [accentColor(_:)](https://developer.apple.com/documentation/swiftui/color/accentcolor(_:)). This should also eliminate the warning issued by Xcode regarding `Accent color 'AccentColor' is not present in any asset catalogs.` Otherwise if PrimaryColor is preferred, then we should update `[AzureCalling|Pods].xcodeproj`to point the necessary accent fields to PrimaryColor.

I should have switched all accentColor into primaryColor but looks like there are two places not being commit. Updated. 
It's better to make the color's name consistent with Figma design

palatter (dismissed):

Gemakk (dismissed):

AzureCalling/Controllers/HangupConfirmationViewController.swift
```diff
@@ -19,12 +19,16 @@ class HangupConfirmationViewController: UIViewController {
         view.backgroundColor = UIColor(red: 0, green: 0, blue: 0, alpha: 0.8)
         view.isOpaque = false
 
+        let primaryColor = UIColor(named: "PrimaryColor") ?? .systemBlue
```
waynemo, line 22: Our color lookups should never be null. So we shouldn't need this conditional.
Could we define some app colors constants instead?

waynemo (changes requested):

AzureCalling/Controllers/HangupConfirmationViewController.swift
AmyL219: Add ColorTheme as a helper class
UIColor(named:) could return nil, provide a fallback value avoid crash

AzureCalling/Utilities/ThemeColor.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import UIKit
+
+enum ThemeColor {
+    static var primary = UIColor(named: "PrimaryColor") ?? UIColor(red: 0, green: 120 / 255, blue: 212 / 255, alpha: 1)
+
+    static var gray300 = UIColor(named: "Gray300") ?? UIColor(red: 172 / 255, green: 172 / 255, blue: 172 / 255, alpha: 1)
```
waynemo, line 12: Let's remove the defaults here. 

In this situation, we are expecting to always have the color in the assets. If we provide a default, it might actually hide the problem if the asset was removed/renamed.

waynemo (changes requested):

AzureCalling/Utilities/ThemeColor.swift
AmyL219: Sounds good. Removed

waynemo (dismissed):

Gemakk (dismissed):

palatter (approved):

Gemakk (approved):

PR #24, opened 2021-04-21
Select output device for audio

Gemakk:
## Purpose

* Show audio device selection menu for Lobby and Call View
* Audio device selection menu shows two options, iPhone and Speaker
* iPhone selection includes any external mic connected (Bluetooth, wired headset)
* Selecting a given audio device will use that device during a call

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test

* Test Audio device selection in Lobby and Call Views
* Connect external audio devices before and during a call

## What to Check
* Audio device menu shows correct and current audio device
* Selection of alternative audio device properly changes audio device used


AzureCalling/External/Calling/AudioDeviceSelectionManager.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import AVFoundation
+
+class AudioDeviceSelectionManager {
```
waynemo, line 9: I think it would be better if the inputs/outputs to this class are purely enum values or IDs instead of a data model. The ViewController can handle mapping to specific images and selected state in the UI.

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+                   didSelectRowAt indexPath: IndexPath) {
+        let cell = tableView.cellForRow(at: indexPath) as! BottomDrawerCellView
+        let audioDeviceDataModel = AudioDeviceDataModel(image: cell.avatar.image!, name: cell.title.text!, enabled: false)
+        audioDeviceSelectionManager.switchAudioDevice(audioDeviceDataModel: audioDeviceDataModel)
+        dismiss(animated: true, completion: nil)
+    }
+
+    // MARK: Actions
+
+    @IBAction func overLayViewDidTapped(_ sender: UITapGestureRecognizer) {
```
waynemo, line 93: Maybe simplify naming to "overlayTapped"

AzureCalling/External/Calling/AudioDeviceDataModel.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+struct AudioDeviceDataModel {
+    let image: UIImage
+    let name: String
+    let enabled: Bool
+}
```
waynemo, line 12: This is not really a data model since it has a UIImage. I think we can call it `AudioDeviceOption` and just define it as part of the `AudioDeviceSelectionViewContoller`.

AzureCalling/External/Calling/TableViewDataSource.swift
```diff
@@ -0,0 +1,54 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class TableViewDataSource: NSObject, UITableViewDataSource {
+
+    // MARK: Properties
+
+    private var cellViewModelDataArray: [BottomDrawerCellViewModel]
```
waynemo, line 12: To keep things simpler, I don't think we need the abstraction of `BottomDrawerCellViewModel`. We can just have an array of "AudioDeviceOption" objects.

AzureCalling/External/Calling/TableViewDataSource.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class TableViewDataSource: NSObject, UITableViewDataSource {
```
waynemo, line 8: I think it would be simpler if we kept this as part of the `AudioDeviceSelectionViewController`.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -131,6 +135,15 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         cleanViewRendering()
     }
 
+    private func openAudioDeviceDrawer() {
+        let audioDeviceSelectionViewController = storyboard?.instantiateViewController(withIdentifier: "AudioDeviceSelectionViewController") as! AudioDeviceSelectionViewController
```
waynemo, line 139: Are we instantiating the viewcontroller from storyboard because the audio selection UI is defined in the Storyboard?

waynemo (changes requested):

AzureCalling/External/Calling/TableViewDataSource.swift
Gemakk: The idea behind his was for it to be compatible with the participant list, which will also make use of BottomDrawerCellViewModel.

AzureCalling/External/Calling/TableViewDataSource.swift
Gemakk: The idea behind was also to be generic enough to be reused for the participant list

AzureCalling/Controllers/CallViewController.swift
Gemakk: It's to make IBOutlets instantiate.  If we just called AudioDeviceSelectionViewController(), the IBOutlet properties would be nil.

AzureCalling/Views/BottomDrawerCellView.swift
```diff
@@ -0,0 +1,28 @@
...
+class BottomDrawerCellView: UITableViewCell {
+
+    // MARK: IBOutlets
+
+    @IBOutlet weak var avatar: UIImageView!
+    @IBOutlet weak var title: UILabel!
+
+    // MARK: UITableViewCell events
+
+    required init?(coder: NSCoder) {
```
AmyL219, line 17: Can we remove the init override func

AzureCalling/Views/BottomDrawerCellView.swift
```diff
@@ -0,0 +1,28 @@
...
+        super.init(coder: coder)
+    }
+
+    // MARK: Public Functions
+
+    public func updateCellView(cellViewModel: BottomDrawerCellViewModel) {
+        self.title.text = cellViewModel.title
+        self.avatar.image = cellViewModel.avatar
+        self.accessoryType = cellViewModel.enabled ? .checkmark : .none
```
AmyL219, line 26: Let's use the checkmark from FluentUI icons

AzureCalling/External/Calling/BottomDrawerCellViewModel.swift
```diff
@@ -0,0 +1,18 @@
...
+import UIKit
+
+struct BottomDrawerCellViewModel {
+    let avatar: UIImage
+    let title: String
+    var enabled: Bool
+}
+
+extension BottomDrawerCellViewModel {
+    func convertToAudioDeviceDataModel() -> AudioDeviceDataModel {
```
AmyL219, line 15: Do we need to convert the viewmodel back to data model?

AzureCalling/External/Calling/AudioDeviceSelectionManager.swift
```diff
@@ -0,0 +1,60 @@
...
+        let route = AVAudioSession.sharedInstance().currentRoute
+        for desc in route.outputs {
+            if desc.portType == .builtInSpeaker {
+                return AudioDeviceDataModel(image: UIImage(named: "ic_fluent_speaker_2_28_filled")!, name: "Speaker", enabled: true)
+            }
+        }
+        return AudioDeviceDataModel(image: UIImage(named: "ic_fluent_speaker_2_28_regular")!, name: "iPhone", enabled: true)
+    }
+
+    public func switchAudioDevice(audioDeviceDataModel: AudioDeviceDataModel) {
```
AmyL219, line 39: let's pass an enum type into the function:
`switchAudioOutput(audioOutputType:)`

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+                                      bundle: nil)
+        deviceDrawer.layer.cornerRadius = 8
+        deviceDrawer.register(audioDeviceCell, forCellReuseIdentifier: "BottomDrawerCellView")
+        deviceDrawer.dataSource = audioDeviceTableDataSource
+        deviceDrawer.delegate = self
+        audioDeviceTableDataSource?.deselectAllRows()
+        let currentAudioDevice = audioDeviceSelectionManager.getCurrentAudioDevice()
+        audioDeviceTableDataSource?.selectRow(title: currentAudioDevice.name)
+        deviceDrawer.reloadData()
+        hideDeviceDrawer()
```
AmyL219, line 41: I think the logic here is not quite clear. Let's classify those code into different categories function like:
-setConstraint
-setTableViewDelegate/Datasource
-setup UI

And why does here need to hide and show drawer when the view getting loaded?

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class AudioDeviceSelectionViewController: UIViewController, UITableViewDelegate {
```
AmyL219, line 8: I don't think we need to hold `audioDeviceSelectionManager` inside the tableViewController. 

This tableview could be reused for participant list drawer, couple audioDeviceSelectionManager inside the tableview doesn't make it reusable for other datasource type.

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    // MARK: Properties
+
+    var audioDeviceSelectionManager: AudioDeviceSelectionManager!
+    var audioDeviceTableDataSource: TableViewDataSource!
```
AmyL219, line 13: Let's move audioDeviceTableDataSource to this class rather than using an object hold it

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        let currentAudioDevice = audioDeviceSelectionManager.getCurrentAudioDevice()
+        audioDeviceTableDataSource?.selectRow(title: currentAudioDevice.name)
+        deviceDrawer.reloadData()
+        hideDeviceDrawer()
+        showDeviceDrawer()
+    }
+
+    // MARK: Private Functions
+
+    private func hideDeviceDrawer() {
```
AmyL219, line 47: For hide action, could we just do view remove rather than constraint change?
``
self.view.removeFromSuperView()

``

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        }, completion: nil)
+    }
+
+    // MARK: UITableViewDelegate
+
+    func tableView(_ tableView: UITableView,
+                   didSelectRowAt indexPath: IndexPath) {
+        let cell = tableView.cellForRow(at: indexPath) as! BottomDrawerCellView
+        let audioDeviceDataModel = AudioDeviceDataModel(image: cell.avatar.image!, name: cell.title.text!, enabled: false)
+        audioDeviceSelectionManager.switchAudioDevice(audioDeviceDataModel: audioDeviceDataModel)
```
AmyL219, line 87: For selected row, we could get the selection datasource by
```
let selectedRow = indexPath.row
let selectedObject = datasource[selecedRow]
```

Could we using a closure here to pass the selectedRow to the view which holds the drawer? Like 



```
let selectedRow = indexPath.row
didSelectedData(selectedRow)
```

And in the CallViewController
```
didSelectedData =  { index in 
  let selected = AudioOutputDataType[index]
  Manager.updatedSelectedDevie(selected)
}

```

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+
+        let audioDeviceCell = UINib(nibName: "BottomDrawerCellView",
+                                      bundle: nil)
+        deviceDrawer.layer.cornerRadius = 8
+        deviceDrawer.register(audioDeviceCell, forCellReuseIdentifier: "BottomDrawerCellView")
+        deviceDrawer.dataSource = audioDeviceTableDataSource
+        deviceDrawer.delegate = self
+        audioDeviceTableDataSource?.deselectAllRows()
+        let currentAudioDevice = audioDeviceSelectionManager.getCurrentAudioDevice()
+        audioDeviceTableDataSource?.selectRow(title: currentAudioDevice.name)
```
AmyL219, line 39: Could `selectRow` be removed?
 Instead, the data should be determined if is selected when the data is being created

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -115,6 +125,15 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
         previewCenterImageView.isHidden = true
     }
 
+    private func openAudioDeviceDrawer() {
+        let audioDeviceSelectionViewController = storyboard?.instantiateViewController(withIdentifier: "AudioDeviceSelectionViewController") as! AudioDeviceSelectionViewController
+        audioDeviceSelectionViewController.audioDeviceSelectionManager = audioDeviceSelectionManager
+        audioDeviceSelectionViewController.audioDeviceTableDataSource = audioDeviceTableDataSource
+        audioDeviceSelectionViewController.modalPresentationStyle = .overCurrentContext
+        audioDeviceSelectionViewController.modalTransitionStyle = .crossDissolve
```
AmyL219, line 133: Could `modalTransitionStyle` and `modalPresentationStyle` be defined in `audioDeviceSelectionViewController`?
Which could avoid duplicated call

AmyL219 (commented):
A few thoughts in code refactoring. Please let me know if there is ant question

AzureCalling/AppDelegate.swift
```diff
@@ -70,7 +70,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         let audioSession = AVAudioSession.sharedInstance()
         do {
-            try audioSession.setCategory(.playAndRecord, mode: .voiceChat)
+            let options: AVAudioSession.CategoryOptions = [.allowBluetooth, .duckOthers, .interruptSpokenAudioAndMixWithOthers, .allowBluetoothA2DP]
```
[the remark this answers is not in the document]
Gemakk: It reduces other app volume while your call is in session.  Documentation here: https://developer.apple.com/documentation/avfaudio/avaudiosession/categoryoptions/1616618-duckothers

Common use cases include keeping turn-by-turn directions going while in a call.

AzureCalling/External/Calling/AudioDeviceSelectionManager.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+
+import UIKit
+import AVFoundation
+
+class AudioDeviceSelectionManager {
+
+    // MARK: Public Functions
+
+    public func initAudioDevicess() -> [AudioDeviceDataModel] {
```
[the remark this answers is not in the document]
Gemakk: At the moment, our audio override options are either `.none` or `.speaker`.  When set to `.none`, this covers the iPhone earpiece, or any bluetooth speaker/headset, with the bluetooth headset taking priority over the iPhone earpiece.  I see that it might be possible to more specifically select these outputs as they become available, but I would suggest that they be a separate ticket.

AzureCalling/External/Calling/AudioDeviceSelectionManager.swift
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        let iPhoneDevice = AudioDeviceDataModel(image: UIImage(named: "ic_fluent_speaker_2_28_regular")!, name: "iPhone", enabled: false)
+        let speakerPhone = AudioDeviceDataModel(image: UIImage(named: "ic_fluent_speaker_2_28_filled")!, name: "Speaker", enabled: true)
+
+        let audioDevices = [iPhoneDevice, speakerPhone]
+        return audioDevices
+    }
+
+    public func getCurrentAudioDevice() -> AudioDeviceDataModel {
```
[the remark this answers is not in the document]
Gemakk: I think I can remove the initAudioDevices completely, yes.

AzureCalling/External/Calling/BottomDrawerCellViewModel.swift
Gemakk: My refactor should address this

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
Gemakk: We need to implement the initial position such that it will slide up from the bottom.  So perhaps it was mis-named as I wanted to put it below the current view rather than hide it

AzureCalling/Controllers/LobbyViewController.swift
Gemakk: modalPresentationStyle must be defined before presenting, but the modalTransitionStyle can be done after.

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+            self.view.layoutIfNeeded()
+        }, completion: nil)
+    }
+
+    func createAudioDeviceOptions() {
+        let audioDeviceTypes = AudioSessionManager.getAllAudioDeviceTypes()
+        let currentAudioDeviceType = AudioSessionManager.getCurrentAudioDeviceType()
+
+        for audioDeviceType in audioDeviceTypes {
+            let audioDeviceOption = AudioDeviceOption(image: audioDeviceType.image, name: audioDeviceType.name, accessoryImage: audioDeviceType.accessoryImage, enabled: audioDeviceType == currentAudioDeviceType)
```
AmyL219, line 94: Since there is an output model from AudioSessionManager already, would it be necessary to convert the `audioDeviceType` to `AudioDeviceOption` then to `bottonCellViewData`? Could it be `audioDeviceType` to `bottomCellViewData`? which we could delete `audioDeviceOption` data type

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // MARK: UITableViewDelegate events
+
+    func tableView(_ tableView: UITableView,
+                   didSelectRowAt indexPath: IndexPath) {
+        let cell = tableView.cellForRow(at: indexPath) as! BottomDrawerCellView
+        let audioDeviceType = AudioDeviceType(rawValue: cell.title.text)!
+        AudioSessionManager.switchAudioDeviceType(audioDeviceType: audioDeviceType)
+        dismissSelf(sender: self)
```
AmyL219, line 106: For `didSelectRowAt` at tableView. For the selected data source, we usually get access it by 
```
let selectedItem = audioDeviceOptions[indexPath]
```
rather than get the selected data by using the cellView. 
See example:
https://developer.apple.com/documentation/uikit/uitableviewdelegate/handling_row_selection_in_a_table_view?language=objc

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+                                      bundle: nil)
+        deviceTable.register(audioDeviceCell, forCellReuseIdentifier: "BottomDrawerCellView")
+        deviceTable.dataSource = self
+        deviceTable.delegate = self
+        deviceTable.reloadData()
+
+        var deviceTableConstraints = [
+        deviceTable.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+        deviceTable.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+        deviceTable.heightAnchor.constraint(equalToConstant: 100)
```
AmyL219, line 62: we should separate a ticket to handle the dynamic height of the table view. And make the table view reusable for `participant view list`. 
We could call the audioSessionManager in the ViewController and pass the `cellViewData` into this table view. And then based on closure to pass the selectedData back to viewController. And this tableViewController has no idea what type of data it's displaying

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+        hideConstraint.priority = .defaultLow
+        deviceTableConstraints.append(hideConstraint)
+
+        NSLayoutConstraint.activate(deviceTableConstraints)
+    }
+
+    func openDeviceTable() {
+        deviceTable.isHidden = false
+
+        let centerYConstraint = NSLayoutConstraint(item: deviceTable!,
```
AmyL219, line 75: `CenterYConstraint` means on the view centre. But here is not in centre, instead is bottom align. We could rename this constraint

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+        return audioDeviceOptions.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: "BottomDrawerCellView", for: indexPath) as? BottomDrawerCellView
+        let audioOption = audioDeviceOptions[indexPath.row]
+        let bottomDrawerCellViewModel = BottomDrawerCellViewModel(avatar: audioOption.image, title: audioOption.name, accessoryImage: audioOption.accessoryImage, enabled: audioOption.enabled)
+        cell?.updateCellView(cellViewModel: bottomDrawerCellViewModel)
+
+        return cell ?? UITableViewCell()
```
AmyL219, line 121: To remove the `??` mark, you need a force unwrap on the cell dequeue:
 ```
        let cell = tableView.dequeueReusableCell(withIdentifier: "BottomDrawerCellView", for: indexPath) as! BottomDrawerCellView
```

File: AzureCalling/Controllers/AudioDeviceSelectionViewController.swift [hunk 26]
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+}
+
+struct AudioDeviceOption {
```
AmyL219, line 126: We could remove this model

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -115,6 +116,12 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
         previewCenterImageView.isHidden = true
     }
 
+    private func openAudioDeviceDrawer() {
+        let audioDeviceSelectionViewController = AudioDeviceSelectionViewController()
+        audioDeviceSelectionViewController.modalPresentationStyle = .overCurrentContext
```
AmyL219, line 121: Could `audioDeviceSelectionViewController.modalPresentationStyle = .overCurrentContext` remove to the `SelectionViewController`?

AzureCalling/External/Calling/AudioSessionManager.swift
```diff
@@ -0,0 +1,47 @@
...
+
+import UIKit
+import AVFoundation
+
+class AudioSessionManager {
+
+    // MARK: Public Functions
+
+    public static func getAllAudioDeviceTypes() -> [AudioDeviceType] {
+        let audioDevices = [AudioDeviceType.receiver, AudioDeviceType.speaker]
```
AmyL219, line 14: Could `AudioDeviceType` be removed when we indicate the type? Something like:
```
let audioDevice:[AudioDeviceType] = [.receiver, speaker]
```

AzureCalling/External/Calling/AudioSessionManager.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public static func getCurrentAudioDeviceType() -> AudioDeviceType {
+        let route = AVAudioSession.sharedInstance().currentRoute
+        for desc in route.outputs {
+            if desc.portType == .builtInSpeaker {
+                return AudioDeviceType.speaker
+            }
+        }
+        return AudioDeviceType.receiver
```
AmyL219, line 25: Could AudioDeviceType be removed?

AmyL219 (commented):
The new code is much clean and easy for understand. 
Just a few more comments on this big PR.

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
Gemakk: Right, good catch!  I previously had BottomCellViewDataModel have different properties than AudioDeviceOption, but I made them the same at same point.  Removing

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
Gemakk: Done.  Modified it such that the table height is dynamic to its content size, and I also include proper margin to account for different iPhone curvatures.  Here's an iPhone SE:
image
And iPhone 12 Pro:
image

AmyL219 (commented):
Just walk through the code locally and do some tests. Found the tableview's checkmark doesn't move to `iphone speaker` after I tap it. Is it because of it's from simulator?

AzureCalling/External/Calling/AudioDeviceType.swift
```diff
@@ -0,0 +1,29 @@
...
+enum AudioDeviceType: String {
+    case receiver = "iPhone"
+    case speaker = "Speaker"
+
+    var image: UIImage {
+        switch self {
+        case .receiver:
+            return UIImage(named: "ic_fluent_speaker_2_28_regular")!
+        case .speaker:
+            return UIImage(named: "ic_fluent_speaker_2_28_filled")!
```
cheukchuen, line 18: Just a thought on whether we should set the filled image based on selection instead of device type. For example, receiver is filled and speaker is regular if receiver is selected, whereas speaker is filled and receiver is regular if speaker is selected

AzureCalling/External/Calling/BottomDrawerCellViewModel.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+struct BottomDrawerCellViewModel {
```
waynemo, line 8: This is not really a view model since it contains references to UI classes. Perhaps we can rename it to something like `BottomDrawerItem`.

waynemo (dismissed):

AzureCalling/External/Calling/BottomDrawerCellViewModel.swift
AmyL219: Yukon wanted to rename the class too.  My original thought of naming the class as `ViewModel` because it's [a common usage for tableview cell](https://cocoacasts.com/simplifying-table-views-with-model-view-viewmodel). 
Whatever, I am fine with renaming it

AzureCalling/External/Calling/AudioDeviceType.swift
Gemakk: This can be added in the future if choose to change the icon base on whether it's enabled or not.  We'll need to think a bit more about it however, since I deliberately did not include state in this enum.

AzureCalling/External/Calling/AudioDeviceType.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+import UIKit
+
+enum AudioDeviceType: String {
+    case receiver = "iPhone"
```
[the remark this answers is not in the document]
Gemakk: Ok; in this case, we'd probably need the View Controller to decide what to return instead of the enum (The enum can return just `receiver`).  I'd suggest this be a future ticket, since detection of the iOS device type has not been done yet.

AzureCalling.xcodeproj/project.pbxproj
```diff
@@ -298,6 +316,7 @@
 			files = (
 				1F7CD3FB25C1E9E80081D8F1 /* LaunchScreen.storyboard in Resources */,
 				1F7CD40325C1E9E80081D8F1 /* AppSettings.plist in Resources */,
+				8D743158263373ED000D0BCE /* BottomDrawerCellView.xib in Resources */,
```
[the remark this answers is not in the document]
Gemakk: image
This is how it looks like for large and small screens.  I found that only setting the autoresizing settings positions the elements correctly.

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.modalTransitionStyle = .crossDissolve
+
+        view.backgroundColor = UIColor(red: 0, green: 0, blue: 0, alpha: 0.8)
+        view.isOpaque = false
+
+        // tapping anywhere on the view is the same as tapping cancel
+        let tap = UITapGestureRecognizer(target: self, action: #selector(dismissSelf))
+        tap.cancelsTouchesInView = false
+        view.addGestureRecognizer(tap)
+        view.isUserInteractionEnabled = true
```
AmyL219, line 28: view.isUserInteractionEnabled default value is true. We could remove this line
Do we need `tap.cancelsTouchesInView`?

AmyL219:
Passed a brief test

AmyL219 (dismissed):

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+        deviceTable.reloadData()
+
+        let window = UIApplication.shared.windows[0]
+        let guide = view.safeAreaLayoutGuide
+        let bottomPadding = window.safeAreaInsets.bottom
+
+        var deviceTableConstraints = [
+            deviceTable.leadingAnchor.constraint(equalTo: guide.leadingAnchor),
+            deviceTable.trailingAnchor.constraint(equalTo: guide.trailingAnchor),
+            deviceTable.heightAnchor.constraint(equalToConstant: deviceTable.contentSize.height + bottomPadding)
```
cheukchuen, line 63: Found an edge case issue with the `heightAnchor` using a constant value while I'm working on the participant list. I know it's highly unlikely that we have large amount of audio device options to choose from, but the table doesn't display large list properly.

I have 2 suggestions on this:
1. If we can guarantee that the number of devices is small and still want the table to cover the full screen, then set `heightAnchor` with the argument `equalTo: guide.heightAnchor` instead of using `equalToConstant`. Otherwise, see below.
2. Use `topAnchor.constraint(equalTo: guide.centerYAnchor)` instead of `heightAnchor`, and set `isScrollEnabled` to true. This will set the table view to occupy half the screen and will have the same behavior as displaying the participant list.

Feel free to provide other suggestions if necessary.

cheukchuen (changes requested):

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
Gemakk: Since the Audio Device list is always static in this context, we can keep it the way it is without issue.  For participant list, we should probably use dynamic to content size up to a certain maximum size, and then make it scrollable, such as your 2nd suggestion.  So I agree it should be done as part of its refactoring to a common class and extension for participant list, but not for this particular ticket.

cheukchuen: For sure. We could do the refactoring separately if that's more appropriate.

AmyL219 (approved):

cheukchuen (approved):

PR #27, opened 2021-04-22
Display indicator on portrait for active speakers

cheukchuen:
## Purpose

* Display indicator on portrait for active speakers

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Start or join a call and check for indicator when remote participants speak

## What to Check
* 1:1 group call with a single border when the other person speaks
* group call with 3+ participants showing multiple borders while they speak
* no border when they are muted


AzureCalling/Views/ParticipantView.swift
```diff
@@ -45,6 +46,10 @@ class ParticipantView: UIView {
         participantLabelView.updateMuteIndicator(isMuted: isMuted)
     }
 
+    func updateActiveSpeaker(isSpeaking: Bool) {
+        activeSpeakerView.isHidden = !isSpeaking
```
AmyL219, line 50: Inside of adding an activeSpeakerViewLayer, could we just update the participantView.border and borderColor?

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -361,6 +361,7 @@ class CallingContext: NSObject {
                     }
                 }
             }
+            self.notifyRemoteParticipantIsSpeakingChanged()
```
AmyL219, line 364: 1. I think `notifyRemoteParticipantIsSpeakingChanged` function is completely overlap with `notifyRemoteParticipantIsMutedChanged`, they both notify the view update. Should we combine them into one function like `notifyRemoteParticipantViewChanged`?
2. on the close of `onIsSpeakingChanged`, there already has some view updating business logic. I think we could update the logic to optimize the notification? Something similar to 
 -If the participant is speaking and not in the grid update the view(which's code already exist in the project) 
-otherwise if participant is in the grid, update the view 
-otherwise no update

cheukchuen: 1. True. Initially I thought we could provide optimization on targeted individual updates in the future, but we could also leave the need for separate notifications (if necessary) in the future as well.
2. Good point. We did that with `onIsMutedChanged` as well so it's logical to bring that optimization over.

AzureCalling/Views/ParticipantView.swift
cheukchuen: I guess this one confuses me as to whether we manipulate UI logic in code or rely on the views in storyboard

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -344,21 +344,23 @@ class CallingContext: NSObject {
...
                 return
             }
-            if remoteParticipant.isSpeaking,
-               let userIdentifier = remoteParticipant.identifier.stringValue {
-                // Swap in speaking participant if not currently displayed
+            if let userIdentifier = remoteParticipant.identifier.stringValue {
                 if self.displayedRemoteParticipants.value(forKey: userIdentifier) == nil {
-                    if self.displayedRemoteParticipants.count == CallingContext.remoteParticipantsDisplayed {
+                    // Swap in speaking participant if not currently displayed
+                    if remoteParticipant.isSpeaking,
```
Gemakk, line 358: Would it make sense to combine these conditionals?
``` swift
if self.displayedRemoteParticipants.value(forKey: userIdentifier) == nil, 
  if remoteParticipant.isSpeaking {
    ...
}
```

Gemakk (dismissed):
Tested app

AzureCalling/External/Calling/CallingContext.swift
cheukchuen: If we combine the conditionals in the `if` statements, we'll have to add another condition in the `else` statement:
```swift
if self.displayedRemoteParticipants.value(forKey: userIdentifier) == nil,
   remoteParticipant.isSpeaking,
   self.displayedRemoteParticipants.count == CallingContext.remoteParticipantsDisplayed {
    ...
} else if self.displayedRemoteParticipants.value(forKey: userIdentifier) != nil {
    ...
}
```
Either way works I guess, it's more of a preference among the team.

AmyL219: I think the `else if` one has better readability for me

AmyL219: Beside this, other code looks good to me

Gemakk (approved):

AmyL219 (approved):

PR #32, opened 2021-05-05
Show static participant list

cheukchuen:
## Purpose

* Show static participant list through the people button on the top banner message

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## What to Check
* Single local participant should show one row
* Multiple participants should grow the list
* List stops at mid screen, which then allows for scrolling
* Changes to the participants (join, leave, mute, unmute) should update the list by dismissing and reopen the list
* Bottom drawer list should function the same for selecting audio devices


Gemakk (dismissed):
Tested with many remote participants.
Just one minor suggestion: Expand the Label maximum size to use up all available space before the checkmark for long names:
image

AzureCalling/Controllers/BottomDrawerViewController.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+protocol BottomDrawerViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
```
AmyL219, line 8: Protocol is usually defined for an interface but not the some generation method. It doesn't contain the implement of the things. I dont think it is the best choice to define BottomDrawerViewController as a protocol even it work here. At least we should define it as a base viewcontroller, and participant list and audio selection drawer can inherit from it. 
Ideally, could the bottomDrawerViewController be the only drawer, and we input different datasource of BottomDrawerItem into the drawer? Maybe the bottomDrawerViewController has 2 different style type, one is for AudioSelection, one is for ParticipantList to build the view difference

AzureCalling/Controllers/ParticipantListViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+
+import UIKit
+
+class ParticipantListViewController: UIViewController, BottomDrawerViewController {
+
+    // MARK: Properties
+
+    private var participantList: [BottomDrawerItem] = [BottomDrawerItem]()
+    var participantTable: UITableView!
+    var callingContext: CallingContext!
```
AmyL219, line 14: I don't think we need to inject the whole callingContext into this class. We just need the remote participant list

Gemakk: We also need our own name and mute status.  Once we implement the live participant list, this will expand to an observer to our own mute status as well and possibly our own speaking indicator.

cheukchuen: Defined a struct `ParticipantListInfo` to carry the necessary info to build the participant list, no longer rely on the entire `callingContext`

AmyL219 (commented):
I think we're in the right page. Just some comments for refactoring suggestion

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
```diff
@@ -5,85 +5,13 @@
 
 import UIKit
 
-class AudioDeviceSelectionViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
+class AudioDeviceSelectionViewController: NSObject, UITableViewDelegate, UITableViewDataSource {
```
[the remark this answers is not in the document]
cheukchuen: Good point. It is also a Delegate though, but I'm fine with DataSource. Where do we place these files though since they are not controllers anymore?

AzureCalling/Controllers/BottomDrawerViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+
+import UIKit
+
+class BottomDrawerViewController: UIViewController {
+
+    // MARK: Properties
+
+    weak var tableViewDelegate: UITableViewDelegate?
+    var tableViewDataSource: UITableViewDataSource?
+    var tableView: UITableView!
```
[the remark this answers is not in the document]
cheukchuen: Agree with the initialization, just that the properties still need to be optional since we also need to add the following init:
```swift
required init?(coder aDecoder: NSCoder) {
    super.init(coder: aDecoder)
}
```

AzureCalling/Controllers/AudioDeviceSelectionViewController.swift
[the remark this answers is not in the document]
AmyL219: I think we could create a new sub folder named `datasource` inside `Controllers`

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -134,9 +135,23 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
 
     private func openAudioDeviceDrawer() {
+        let bottomDrawerViewController = BottomDrawerViewController()
         let audioDeviceSelectionViewController = AudioDeviceSelectionViewController()
-        audioDeviceSelectionViewController.modalPresentationStyle = .overCurrentContext
-        present(audioDeviceSelectionViewController, animated: false, completion: nil)
+        audioDeviceSelectionViewController.createAudioDeviceOptions()
+        bottomDrawerViewController.tableViewDataSource = audioDeviceSelectionViewController
+        bottomDrawerViewController.tableViewDelegate = audioDeviceSelectionViewController
+        bottomDrawerViewController.modalPresentationStyle = .overCurrentContext
```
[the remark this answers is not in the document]
cheukchuen: It works!

AzureCalling/Controllers/ParticipantListViewController.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import AzureCommunicationCalling
+
+struct ParticipantListInfo {
+    let localDisplayName: String
+    let localIsMuted: Bool?
+    let remoteParticipants: MappedSequence<String, RemoteParticipant>
```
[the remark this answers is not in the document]
cheukchuen: Replaced with `[RemoteParticipantInfo]` instead

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -134,9 +135,31 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+            delegate: audioDeviceSelectionViewController)
+        present(bottomDrawerViewController, animated: false, completion: nil)
+    }
+
+    private func openParticipantListDrawer() {
+        let participantListViewController = ParticipantListDataSource()
+        let participantListInfo = ParticipantListInfo(
+            localDisplayName: callingContext.displayName,
+            localIsMuted: callingContext.callIsMuted,
+            remoteParticipants: callingContext.remoteParticipants.map { (remoteParticipant) -> RemoteParticipantInfo in
```
AmyL219, line 152: It would be clear if the converting remoteParticipant and ListInfo are separated. 
```let remoteParticipantInfoList = callingContext.remoteParticipants.map .....
let participantListInfo = ParticipantListInfo(
            localDisplayName: callingContext.displayName,
            localIsMuted: callingContext.callIsMuted,
            remoteParticipants:remoteParticipantInfoList

File: AzureCalling.xcodeproj/project.pbxproj [hunk 8]
```diff
@@ -158,13 +162,14 @@
 		1F7CD3EA25C1E9E80081D8F1 /* Controllers */ = {
 			isa = PBXGroup;
 			children = (
+				989F61952649D72F000C5668 /* datasource */,
 				1F7CD3EB25C1E9E80081D8F1 /* IntroViewController.swift */,
```
AmyL219, line 166: Let's use upper case `DataSource` to keep consistent with other groups

AmyL219 (commented):
Two minor comments. Others looks good to me

AmyL219 (dismissed):

AzureCalling/Controllers/datasource/ParticipantListDataSource.swift
```diff
@@ -0,0 +1,53 @@
...
+struct ParticipantListInfo {
+    let localDisplayName: String
+    let localIsMuted: Bool?
+    let remoteParticipants: [RemoteParticipantInfo]
+}
+
+struct RemoteParticipantInfo {
+    let displayName: String
+    let isMuted: Bool
+}
```
waynemo, line 17: Couldn't we just have a list of `ParticipantInfo` that can be used for both local and remote?

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -37,6 +37,10 @@ class CallingContext: NSObject {
         return remoteParticipantCount + 1
     }
 
+    var callIsMuted: Bool? {
```
waynemo, line 40: I think we should use "is" as the prefix -> `isCallMuted`

waynemo (dismissed):

AzureCalling/Controllers/datasource/ParticipantListDataSource.swift
cheukchuen: Good point! That should simplify and generalize the participant list

AmyL219 (approved):

Gemakk (approved):

Gemakk (approved):
Tested

PR #33, opened 2021-05-05
Overlay border view to indicate speaking

cheukchuen:
## Purpose

* Fixes bug on border being shown partially if we were to change the border programmatically on the content view
* Video rendering causes the border to be interfered with the content, thus overlay remedies this issue by displaying the border on top of the content

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Solid border when speaking indicator is up
* Width of the surrounding border should be consistent and span 3px


AmyL219:
Do you have idea why does the border in participant view miss a part?

Gemakk:
I didn't run into the issue while testing, and in theory, this should solve it, but since the original issue was difficult to reproduce, were you able to reliably reproduce it, Cheuk?

cheukchuen:
> Do you have idea why does the border in participant view miss a part? If it is because the border mix the content view, could `self.layer.masksToBounds = true` fix it ?

@AmyL219 I'm guessing it's trying to render the video along the border. Tried applying `self.layer.masksToBounds = true`, doesn't fix the problem though. Maybe we need to modify the `self.layer.bounds` area to only render within the bounds as well, but this might be cumbersome.

> I didn't run into the issue while testing, and in theory, this should solve it, but since the original issue was difficult to reproduce, were you able to reliably reproduce it, Cheuk?

@Gemakk I'm able to reproduce the bug on partial border and the overlay approach seems to solve it. Couldn't reproduce the bug on larger left margin though. Are you able to see any difference on the margin with the overlay?

AmyL219 (approved):

Gemakk (approved):
I managed to both reproduce the issue in the main branch and was unable to reproduce it in this one.

PR #37, opened 2021-05-14
Update participant list to be dynamic

cheukchuen:
## Purpose

* Allow live update on participant list whenever remote participants join / leave call or mute / unmute

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## What to Check
* Have remote participants join call, that should increase the participant list while the participant list is in view
* Have remote participants leave call, that should decrease the participant list while the participant list is in view
* Have remote participants mute themselves, that should display the mute indicators beside their names while the participant list is in view
* Have remote participants unmute themselves, that should hide the mute indicators beside their names while the participant list is in view
* Have remote participants that are currently in active display mode to switch on / off their video, that should display their video in the background while the participant list is in view
* Have remote participants that are currently in active display mode to unmute themselves and speak, that should display the active speaker indicators around their portraits in the background while the participant list is in view

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -18,6 +18,8 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
     var joinCallConfig: JoinCallConfig!
     var callingContext: CallingContext!
 
+    private var bottomDrawerViewController: BottomDrawerViewController?
+
     private let eventHandlingQueue = DispatchQueue(label: "eventHandlingQueue", qos: .userInteractive)
     private var lastParticipantViewsUpdateTimestamp: TimeInterval = Date().timeIntervalSince1970
     private var isParticipantViewsUpdatePending: Bool = false
@@ -27,23 +29,6 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
     private var localParticipantView = ParticipantView()
     private var participantIdIndexPathMap: [String: IndexPath] = [:]
     private var participantIndexPathViewMap: [IndexPath: ParticipantView] = [:]
-
-    private var bottomDrawerViewController: BottomDrawerViewController?
-    private var participantInfoList: [ParticipantInfo] {
-        // Show local participant first
-        var participantInfoList = [
-            ParticipantInfo(
-                displayName: callingContext.displayName + " (Me)",
-                isMuted: callingContext.isCallMuted ?? false)
-        ]
-        // Get the rest of remote participants
-        participantInfoList.append(contentsOf: callingContext.remoteParticipants.map {
-            ParticipantInfo(
-                displayName: $0.displayName,
-                isMuted: $0.isMuted)
-        })
-        return participantInfoList
-    }
 
     // MARK: IBOutlets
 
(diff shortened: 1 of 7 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -578,13 +585,30 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+            guard let self = self,
+                  let bottomDrawerViewController = self.bottomDrawerViewController,
+                  bottomDrawerViewController.isViewLoaded else {
+                return
+            }
+
+            let participantListDataSource = ParticipantListDataSource()
+            let participantInfoList = self.getParticipantInfoList()
+            participantListDataSource.createParticipantList(participantInfoList)
+            bottomDrawerViewController.refreshBottomDrawer(dataSource: participantListDataSource)
```
AmyL219, line 599: I wonder for the class of DataSource type, could it be possible to input the participantInfoList as an initialization parameter/closure (like tokenFetchClosure) input to them? Reason why I think so is, the participantListDataSource is useless until there is participantInfoList/Device data input.

AmyL219 (commented):
Just one comment for possible refactoring

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -578,13 +585,30 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+        }
+    }
+
     @objc func onRemoteParticipantsUpdated(_ notification: Notification) {
         queueParticipantViewsUpdate()
+        participantListUpdate()
```
Gemakk, line 605: Have you considered whether or not it made sense to put this method into `queueParticipantViewsUpdate`?

cheukchuen: I thought about it before, but decided to separate them into their respective methods since they both handle different parts of the updates: one is to update specifically the active participants view, another to just update the list. Separation of concerns in a way.

There's also the delayed update mechanism for `updateParticipantViews`, unless we want the delay to happen when updating the participant list too, in which case combining both updates might make sense.

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Good point. Changed to use closure for the init. Thanks for the suggestion!

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -145,7 +147,16 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
     }
 
     private func openParticipantListDrawer() {
-        let participantListDataSource = ParticipantListDataSource()
+        let participantListDataSource = ParticipantListDataSource(participantsFetcher: getParticipantInfoList)
+        participantListDataSource.createParticipantList()
```
AmyL219, line 151: Could`createParticipantList()` also been moved into ParticipantListDataSource() class?

cheukchuen: Moved `createParticipantList()` to be invoked in the inits. Also modified the refresh mechanism since we now have the fetcher in the data sources

AzureCalling/Controllers/datasource/ParticipantListDataSource.swift
```diff
@@ -12,20 +12,25 @@ struct ParticipantInfo {
...
 
     private var participantList = [BottomDrawerItem]()
     private var participantsFetcher: ParticipantsFetcher
 
+    // MARK: Initialization
+
     init(participantsFetcher: @escaping ParticipantsFetcher) {
         self.participantsFetcher = participantsFetcher
+
+        super.init()
```
AmyL219, line 27: Why don't we put `super.init()` before assigning participantsFetcher? Is it because alert here?

AmyL219 (approved):
Code looks cleaner for me

AzureCalling/Controllers/datasource/ParticipantListDataSource.swift
cheukchuen: It could be that `participantsFetcher` is non-optional, hence it needs to be initialized prior to calling `super.init()`

Gemakk (approved):
Tested

PR #39, opened 2021-05-18
Add logic to spotlight screen share

jimchou-dev:
## Purpose
* Added logic to spotlight screen share

```
[ ] Yes
[ x ] No
```

## Pull Request Type
```
[ ] Bugfix
[ x ] Feature
[ ] Other... Please describe:
```

## What to Check
* If If screen share is present, it will show picture-in-picture view focusing on the person doing the screen share
* Screen share will `Fit` to view instead of `Crop` to view

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -448,6 +448,43 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         }
     }
 
+    private func getOrCreateParticipantView(participant: RemoteParticipant,
+                                            index: Int,
+                                            prevParticipantIdIndexPathMap: [String: IndexPath],
+                                            prevParticipantIndexPathViewMap: inout [IndexPath: ParticipantView],
+                                            indexPathMoves: inout [(at: IndexPath, to: IndexPath)],
+                                            insertIndexPaths: inout [IndexPath]) {
+        let userIdentifier = participant.identifier.stringValue ?? ""
+        let indexPath = IndexPath(item: index, section: 0)
+
+        var participantView: ParticipantView?
+        if let prevIndexPath = prevParticipantIdIndexPathMap[userIdentifier],
+           let prevParticipantView = prevParticipantIndexPathViewMap[prevIndexPath] {
+            prevParticipantIndexPathViewMap.removeValue(forKey: prevIndexPath)
+            participantView = prevParticipantView
+            if prevIndexPath != indexPath {
+                // Add to move list
+                indexPathMoves.append((at: prevIndexPath, to: indexPath))
+            }
+        } else {
+            participantView = ParticipantView()
+            // Add to insert list
+            insertIndexPaths.append(indexPath)
+        }
+
+        participantView!.updateDisplayName(displayName: participant.displayName)
+        participantView!.updateMuteIndicator(isMuted: participant.isMuted)
+        participantView!.updateActiveSpeaker(isSpeaking: participant.isSpeaking)
+        if let videoStream = participant.videoStreams.first(where: { $0.mediaStreamType == .screenSharing }) {
+            participantView!.updateVideoStream(remoteVideoStream: videoStream, isScreenSharing: true)
+        } else {
+            participantView!.updateVideoStream(remoteVideoStream: participant.videoStreams.first)
+        }
+
+        participantIdIndexPathMap[userIdentifier] = indexPath
+        participantIndexPathViewMap[indexPath] = participantView!
+    }
+
     private func updateParticipantViews(completionHandler: @escaping () -> Void) {
         // Previous maps tracking participants
         let prevParticipantIdIndexPathMap = participantIdIndexPathMap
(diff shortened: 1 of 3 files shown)


AzureCalling/Views/ParticipantView.swift
```diff
@@ -15,6 +15,7 @@ class ParticipantView: UIView {
     private var renderer: VideoStreamRenderer?
     private var rendererView: RendererView?
     private var videoStreamId: String?
+    public var hasScreenShare: Bool = false
```
AmyL219, line 18: I think this property should be renamed to `isScreenShare`
it should have a private setter. You could assign this property when there is videoStream Pass into. One example: 
```
func updateVideoStream(remoteVideoStream: RemoteVideoStream?) {
   self.isScreenShare = false 
}

func updateSharedScreenVideo(remoteVideoStream: RemoteVideoStream?) {
   self.isScreenShare = true 
}

....
}

File: AzureCalling/Controllers/CallViewController.swift [hunk 2]
```diff
@@ -462,37 +462,76 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
-
-            // Check for previously tracked participants
-            if let prevIndexPath = prevParticipantIdIndexPathMap[userIdentifier],
-               let prevParticipantView = prevParticipantIndexPathViewMap[prevIndexPath] {
-                prevParticipantIndexPathViewMap.removeValue(forKey: prevIndexPath)
+        // Check to see if any screenshare
+        var screenShareParticipantView: ParticipantView?
+        outerLoop: for participant in callingContext.displayedRemoteParticipants {
+            for videoStream in participant.videoStreams {
+                if videoStream.mediaStreamType == .screenSharing {
```
AmyL219, line 469: The following part of code is kind of redundant with the line 504-533, could it be possible to abstract one function with different parameters in put to build the participant view?

AmyL219 (commented):
A brief review with some thoughts.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -462,37 +499,28 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
-                    // Add to move list
-                    indexPathMoves.append((at: prevIndexPath, to: indexPath))
-                }
-            } else {
-                participantView = ParticipantView()
-
-                // Add to insert list
-                insertIndexPaths.append(indexPath)
+        // Check to see if any screenshare
+        for participant in callingContext.displayedRemoteParticipants {
```
AmyL219, line 503: Is that possible replace the callingContext.displayedRemoteParticiopants with a local cache like callingViewController.wantToDisplayedRemoteParticipant. When there is a screensharing user we just put the user into the local cache, otherwise, the localCache would contain the callContext,displayedRemoteParticipants.

And when we for loop, we for loop the local cache callingViewController.wantToDisplayedRemoteParticipant to update the participant view

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -462,37 +462,76 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
-            let userIdentifier = participant.identifier.stringValue ?? ""
-            let indexPath = IndexPath(item: index, section: 0)
-            var participantView: ParticipantView
-
-            // Check for previously tracked participants
-            if let prevIndexPath = prevParticipantIdIndexPathMap[userIdentifier],
-               let prevParticipantView = prevParticipantIndexPathViewMap[prevIndexPath] {
-                prevParticipantIndexPathViewMap.removeValue(forKey: prevIndexPath)
+        // Check to see if any screenshare
+        var screenShareParticipantView: ParticipantView?
```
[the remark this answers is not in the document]
jimchou-dev: The logic is now in callingContext

AzureCalling/Views/ParticipantView.swift
jimchou-dev: Variable renamed to isScreenSharing

AzureCalling/Controllers/CallViewController.swift
jimchou-dev: The logic is now in callingContext

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -462,37 +462,76 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+        outerLoop: for participant in callingContext.displayedRemoteParticipants {
```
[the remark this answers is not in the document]
jimchou-dev: The logic is now in callingContext

AzureCalling/Controllers/CallViewController.swift
jimchou-dev: Please see new commit that has new logic in calling Context onVideoStreamsUpdated

Gemakk (changes requested):
When the presenter had video on before presenting content, once presentation is over, the presenter's video does not return.

AzureCalling/Controllers/CallViewController.swift
AmyL219: The logic change of filtering out the screensharing user doesn't affect the approach I suggest. 
In the approach I mentioned above. You don't need to abstract any helper class and cache the prevIndexPath map, you only need to update the iterate array on the entry of updateParticipantView

Gemakk:
> When the presenter had video on before presenting content, once presentation is over, the presenter's video does not return.

Met with @jimchou-dev and verified that it just takes a while and eventually appears.

AzureCalling/Controllers/CallViewController.swift
jimchou-dev: Revert to the previous implementation to not have the new class variables in the new commit

AmyL219 (commented):
Looks much better and cleaner. Just two more comments

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -369,10 +370,17 @@ class CallingContext: NSObject {
...
+        participantsEventsAdapter?.onVideoStreamsUpdated = { [weak self] remoteParticipant in
             guard let self = self else {
                 return
             }
+            if remoteParticipant.videoStreams.contains(where: { $0.mediaStreamType == .screenSharing }) {
+                self.currentScreenSharingParticipant = remoteParticipant
+            } else if let userIdentifier = remoteParticipant.identifier.stringValue,
+                      let screenSharingParticipantIdentifier = self.currentScreenSharingParticipant?.identifier.stringValue,
+                      userIdentifier == screenSharingParticipantIdentifier {
+                self.currentScreenSharingParticipant = nil
```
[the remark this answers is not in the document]
jimchou-dev: Added logic in removeRemoteParticipants to set it to nil if the screen share user leaves

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -462,8 +462,17 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         var indexPathMoves: [(at: IndexPath, to: IndexPath)] = []
         var insertIndexPaths: [IndexPath] = []
 
+        var remoteParticipantsToDisplay: MappedSequence<String, RemoteParticipant>
+        if let screenSharingParticipant = callingContext.currentScreenSharingParticipant,
```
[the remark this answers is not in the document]
jimchou-dev: Extracted logic into getRemoteParticipantsToDisplay

AmyL219 (approved):
Code looks good for me
 I could have a test later

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -446,6 +447,18 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
...
+           let userIdentifier = screenSharingParticipant.identifier.stringValue {
+            remoteParticipantsToDisplay = MappedSequence<String, RemoteParticipant>()
+            remoteParticipantsToDisplay.append(forKey: userIdentifier, value: screenSharingParticipant)
+        } else {
+            remoteParticipantsToDisplay = callingContext.displayedRemoteParticipants
+        }
+        return remoteParticipantsToDisplay
+    }
```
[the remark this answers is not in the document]
jimchou-dev: I did consider to do something like that, but MappedSequence does not have a key-val initializer

Gemakk (approved):
Tested

PR #41, opened 2021-05-25
While updating participant views, defer any further updates until the current update completes

cheukchuen:
## Purpose

* ~~This pull request creates a separate update function dedicated for both `onIsMutedChanged` and `onIsSpeakingChanged`.~~
* This pull request modifies the participant views update logic so that whenever an update is in progress, we halt any further update requests until the current update completes. Only then we execute another update.
* This addresses the bug reported on https://github.com/Azure/Communication/issues/300 where the `onIsSpeakingChanged` event is being triggered when remote participants leave the call.
* Otherwise, if both `onIsSpeakingChanged` and `onRemoteParticipantsUpdated` events are fired simultaneously, previous implementation uses `queueParticipantViewsUpdate()` for `onIsSpeakingChanged` and `queueParticipantViewsUpdate()` will be invoked twice; the second execution may lead to the evaluation of the expression below, thereby skipping the actual update. This results in the participant views not updating occasionally when remote participants leave the call.
```
if self.isParticipantViewsUpdatePending {
    return
}
```

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Make sure remote participants have their mute and active speaker indicators updated correctly when changed
* Make sure participant views updated correctly when remote participants leave the call

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -594,6 +594,23 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         infoHeaderView.updateParticipant(count: callingContext.participantCount)
     }
 
+    private func participantViewUpdate() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else {
+                return
+            }
+
+            for participant in self.callingContext.displayedRemoteParticipants {
+                if let userIdentifier = participant.identifier.stringValue,
+                   let indexPath = self.participantIdIndexPathMap[userIdentifier],
+                   let participantView = self.participantIndexPathViewMap[indexPath] {
+                    participantView.updateMuteIndicator(isMuted: participant.isMuted)
+                    participantView.updateActiveSpeaker(isSpeaking: participant.isSpeaking)
+                }
+            }
+        }
+    }
+
     private func participantListUpdate() {
         DispatchQueue.main.async { [weak self] in
             guard let self = self,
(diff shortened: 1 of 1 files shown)


Gemakk (commented):
I was able to reproduce the issue as originally stated and was unable to reproduce it in the fix branch.

About `participantViewUpdate`:
-  I find it a bit confusing to intuitively think about what the method does based off of the name.  I'm not sure if the method fully represents what it does, especially in relation to `updateParticipantViews`. 
- The action based off of the notification seems to make sense however:
  - `onRemoteParticipantsUpdated -> queueParticipantViewsUpdate -> updateParticipantViews`
  - `onRemoteParticipantViewChanged -> participantViewUpdate`
- I wonder if there's a way for the addition and removal of participants code to be separate from the updating of their views

cheukchuen:
> I was able to reproduce the issue as originally stated and was unable to reproduce it in the fix branch.
> 
> About `participantViewUpdate`:
> 
> * I find it a bit confusing to intuitively think about what the method does based off of the name.  I'm not sure if the method fully represents what it does, especially in relation to `updateParticipantViews`.
> * The action based off of the notification seems to make sense however:
>   
>   * `onRemoteParticipantsUpdated -> queueParticipantViewsUpdate -> updateParticipantViews`
>   * `onRemoteParticipantViewChanged -> participantViewUpdate`
> * I wonder if there's a way for the addition and removal of participants code to be separate from the updating of their views

We no longer use this implementation. Had a discussion with @AmyL219 and we have another approach to the problem at hand. Another round of review and test would be much appreciated.

AmyL219 (approved):

Gemakk (approved):
Tested

PR #42, opened 2021-05-25
Implement front and back camera switching

cheukchuen:
## Purpose

* Enable front and back camera switching for local participant

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## What to Check
* Front and back camera switching upon tapping the camera switch button
* The camera switch button should be slightly smaller and in the top right for the detached (PIP) local participant view during 1:1
* The camera switch button should be slightly larger and in the middle right for the standard local participant view for all other calls


AzureCalling/Views/ParticipantView.swift
```diff
@@ -105,9 +108,44 @@ class ParticipantView: UIView {
...
+
+        if isOneOnOne {
+            // Top right (smaller) camera switch for detached local participant view
+            buttonLength = 24
+            verticalConstraint = switchCameraButton.topAnchor.constraint(equalTo: containerView.topAnchor, constant: 4)
+        } else {
+            // Middle right camera switch for standard local participant view
+            buttonLength = 30
+            verticalConstraint = switchCameraButton.centerYAnchor.constraint(equalTo: containerView.centerYAnchor)
+        }
```
jimchou-dev, line 133: The middle right camera is to prevent the camera switch button to be covered by the InfoHeader when the caller is the only user in the call and have camera turned on.

When the call grid is is in PIP the switchCameraButton is constraint to the top-right corner. When call view grid is 2x2 or 3x2, the switchCamera Button should also be on the top-right corner. Probably can do a check if callingContext.displayedRemoteParticipants is not empty.

Gemakk (dismissed):
Functionality fully works as tested.  I noticed specific emulator bugs, but I also verified that they're unrelated to this specific PR.

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -226,6 +226,27 @@ class CallingContext: NSObject {
...
+
+        if let flipCamera = availableCameras[flip] {
+            localVideoStream.switchSource(camera: flipCamera) { error in
+                if error != nil {
+                    print("ERROR: It was not possible to switch camera. \(error!)")
+                    completionHandler(.failure(error!))
+                    return
+                }
+                print("Camera switch successful")
+                completionHandler(.success(()))
```
AmyL219, line 245: if there is no flipCamera available, we still need to send out a completionHandler(.failure)

AzureCalling/Views/ParticipantView.swift
cheukchuen: I've thought about this as well, but decided to keep the button on middle right to be consistent on all views except for PIP. It depends on how we want the design to be.

@Gemakk how did you implement this in Android? Do you think we should keep it top-right corner for 2x2 and 3x2 as well?

AzureCalling/Views/ParticipantView.swift
```diff
@@ -105,9 +108,44 @@ class ParticipantView: UIView {
         }
 
         placeholderImage.isHidden = isDisplayVideo
+        switchCameraButton.isHidden = !isDisplayVideo
         view.isHidden = !isDisplayVideo
     }
 
+    func updateCameraSwitch(isOneOnOne: Bool) {
+        // Skip update if display mode is the same
+        guard detachMode != isOneOnOne else {
```
AmyL219, line 117: Could `detachMode` and `isOneOnOne` be the same name?

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -334,9 +355,15 @@ class CallingContext: NSObject {
         guard let deviceManager = deviceManager else {
             return
         }
-        let camera = deviceManager.cameras[0]
-        localVideoStream = LocalVideoStream(camera: camera)
-        completionHandler(localVideoStream)
+        availableCameras[.front] = deviceManager.cameras.first(where: { $0.cameraFacing == .front })
+        availableCameras[.back] = deviceManager.cameras.first(where: { $0.cameraFacing == .back })
+
+        if let frontCamera = availableCameras[.front] {
```
AmyL219, line 361: Do we need to force the localVideoStream to be the front Camera?

AzureCalling/Views/ParticipantView.swift
Gemakk: Default is right.  Top right only during 1:1

jimchou-dev (dismissed):
Tested, LGTM

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -312,6 +311,10 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         infoHeaderView.toggleDisplay()
     }
 
+    @IBAction func switchCamera(_ sender: UIButton) {
```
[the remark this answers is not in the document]
cheukchuen: This is one way I figured to invoke callingContext's method.
We also need to enable this feature in lobbyView as the next work item.

AzureCalling/Views/ParticipantView.swift
cheukchuen: Sure. I could also use an alias for the parameter: `updateCameraSwitch(isOneOnOne detachMode: Bool)`

AzureCalling/External/Calling/CallingContext.swift
cheukchuen: Yes, it was front by default too. Previous implementation `deviceManager.cameras[0]` gets the front camera.

AzureCalling/External/Calling/CallingContext.swift
cheukchuen: Good point, we still need to handle all cases even though it's unlikely to happen. `.failure` does not accept optional Error value though.

AzureCalling/Controllers/CallViewController.swift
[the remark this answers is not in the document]
AmyL219: Let's use closure/delegate to pass the action from participant view rather than direct linking participant view's button to callingView. 
Please correct me if you have any good reference resource with better approach

Gemakk (dismissed):

AzureCalling/Controllers/CallViewController.swift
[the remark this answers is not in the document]
cheukchuen: Closure is perfectly fine in this scenario.

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -64,6 +64,14 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
         updateToggleVideoButtonState()
 
+        localParticipantView.switchCamera = { [weak self] in
+            guard let self = self else {
+                return
+            }
+
+            self.callingContext.switchCamera { _ in }
+        }
```
AmyL219, line 73: How about the localParticipantView in the grid? Do we need to update the switchCamera closure on the onQueueParticipantViewUpdate() function?

cheukchuen: No need. We have only a single instance of localParticipantView, so once we assign a proper switchCamera closure to localParticipantView it'll be functional regardless of the grid configuration (I've tested it).
Remote participants will still have the empty non-functional closure for switchCamera that does nothing.

AmyL219 (dismissed):

AzureCalling/Views/ParticipantView.swift
```diff
@@ -12,6 +12,9 @@ class ParticipantView: UIView {
 
     // MARK: Properties
 
+    var switchCamera: () -> Void = {}
```
waynemo, line 15: Instead of a property, it might be better to have a method to set the callback:
`setOnSwitchCamera()`

AzureCalling/Views/ParticipantView.swift
```diff
@@ -37,6 +42,12 @@ class ParticipantView: UIView {
         initSubviews()
     }
 
+    // MARK: Actions
+
+    @IBAction func switchCamera(_ sender: UIButton) {
```
waynemo, line 47: Since switch camera functionality is only available for local participants, we should consider a separate class for `LocalParticipantView` that extends `ParticipantView`.

waynemo (changes requested):

AzureCalling/Views/ParticipantView.swift
cheukchuen: Good point! That should abstract away some of the logic (including video toggling) pertaining to local participant activities.

AzureCalling/Views/ParticipantView.swift
cheukchuen: Meaning we prefer private fields to limit access and have a setter method instead? Got it.

Gemakk (approved):
A quick triple tap using a real device leave the camera grey.  For remote participants, they are notified that video is on, but nothing appears (Also grey).
image

Gemakk (changes requested):
See comments above

Gemakk (dismissed):
Tested

AzureCalling/Views/ParticipantView.swift
```diff
@@ -20,9 +20,11 @@ class ParticipantView: UIView {
     // MARK: IBOutlets
 
     @IBOutlet var contentView: UIView!
+    @IBOutlet weak var containerView: UIView!
     @IBOutlet weak var placeholderImage: UIImageView!
     @IBOutlet weak var videoViewContainer: UIView!
     @IBOutlet weak var activeSpeakerView: ActiveSpeakerView!
+    @IBOutlet weak var switchCameraButton: UIRoundedButton!
```
waynemo, line 27: Should these newly added properties only belong to LocalParticipantView?

waynemo (changes requested):

AzureCalling/Views/ParticipantView.swift
cheukchuen: They're referencing outlets in the ParticipantView.xib storyboard. Can't seem to connect those properties if we move them to LocalParticipantView, unless we duplicate the storyboard into LocalParticipantView.xib and remove any camera switch elements in ParticipantView.xib.

@AmyL219 you have any idea how to connect the outlets outside of the File Owner's class without duplicating the storyboard? Or that's the only option to have a clear separation of concerns.

cheukchuen: Programmatically configure the button in LocalParticipantView instead of relying on ParticipantView.xib with the latest commit. That should remove any references to camera switch in ParticipantView

waynemo (approved):

AmyL219 (approved):

PR #46, opened 2021-06-08
Fix issue for removed from meeting without alert

jimchou-dev:
## Purpose
* Deal with when end call and when being removed from a call both call state becomes `disconnected` from SDK
* Add a new state to only show alert when local user gets removed from the meeting by Teams client user

```
[ ] Yes
[ x ] No
```

## Pull Request Type
```
[ x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* When being removed from a Teams meeting by a client, it should show alert
* When leaving a meeting, it should not show alert


waynemo (approved):

cheukchuen (approved):
Functionally tested

PR #51, opened 2021-06-23
fix screenshare not showing on join

jimchou-dev:
## Purpose
* Bug fix for when many participant in a call, screen sharing not displayed for new joined user

```
[ ] Yes
[ x ] No
```

## Pull Request Type
```
[ x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* When many participant in a call and someone is screen sharing, the mobile user that join should see screen sharing

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -322,6 +322,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
         infoHeaderView.toggleDisplay()
         meetingInfoViewUpdate()
+        callingContext.updateScreenSharingParticipant()
         initParticipantViews()
         activityIndicator.stopAnimating()
     }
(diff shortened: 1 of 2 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -322,6 +322,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
         infoHeaderView.toggleDisplay()
         meetingInfoViewUpdate()
+        callingContext.updateScreenSharingParticipant()
```
AmyL219, line 325: Does this work for joining a teams' meeting with screensharing by a link?
I don't think it's a good idea to call the seekScreenSharingParticipant from an external view controller. 
We'd better do the seeking when the call state changed to .connected

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -469,6 +478,10 @@ extension CallingContext: CallDelegate {
             if let userIdentifier = participant.identifier.stringValue {
                 participant.delegate = self.participantsEventsAdapter
                 self.remoteParticipants.append(forKey: userIdentifier, value: participant)
+                if participant.videoStreams.contains(where: { $0.mediaStreamType == .screenSharing }) {
```
AmyL219, line 481: Is this necessary? Would this be duplicated with the function inside `onVideoStreamsUpdated`?

jimchou-dev: This would be duplicate function, but I think because there is logic in removeRemoteParticipants to remove the `currentScreenSharingParticipant` there should also be logic to add it when addRemoteParticipants.

But I will remove the code here and let `onVideoStreamsUpdated` handle to check for screen sharing

AzureCalling/Controllers/CallViewController.swift
jimchou-dev: Good call, I have move it to state change .connected.

And I tested the code locally on teams_interop branch and it works for Teams meeting screen share.

AmyL219 (approved):

cheukchuen (approved):
Functionally tested. Takes few seconds to load though

PR #52, opened 2021-06-23
Reset video renderer when backtracking to lobby

cheukchuen:
## Purpose

* This PR fixes the bug when a person turns on the video in the lobby screen and tries to join a teams call, but then denied by the host while in the waiting room and revert back to the lobby screen with the video not visible even though it's on.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Steps to Test
1. Join a Teams meeting
2. In the Lobby screen enable camera
3. When Teams client rejects the Admit request, the mobile app will prompt Join request not allowed
4. When go back to the Lobby screen, the camera feed should show video stream, along with Camera icon enabled
5. Make sure turning the video off/on and camera switching works as well before/after joining a call

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -96,6 +96,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
+        localParticipantView.dispose()
         self.navigationController?.setNavigationBarHidden(false, animated: animated)
         UIApplication.shared.isIdleTimerDisabled = false
         forcePortraitOrientation()
--- a/AzureCalling/Controllers/LobbyViewController.swift
+++ b/AzureCalling/Controllers/LobbyViewController.swift
@@ -91,9 +91,8 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
         NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillHide), name: UIResponder.keyboardWillHideNotification, object: nil)
     }
 
-    private func setupPreviewView(localVideoStream: LocalVideoStream, retry: Int = 0) {
-        if rendererView == nil,
-           retry <= 3 {
+    private func setupPreviewView(localVideoStream: LocalVideoStream) {
+        if rendererView == nil {
             do {
                 loadingView.startAnimating()
                 previewRenderer = try VideoStreamRenderer(localVideoStream: localVideoStream)
@@ -105,17 +104,8 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
                 rendererView!.leftAnchor.constraint(equalTo: previewView.leftAnchor).isActive = true
                 rendererView!.rightAnchor.constraint(equalTo: previewView.rightAnchor).isActive = true
                 loadingView.stopAnimating()
-                print("Lobby video view renderer successfully created.")
             } catch {
-                print("Failed to create renderer for lobby video view. " + (retry > 0 ? "Retry attempt #\(retry)" : "Attempt to retry..."))
-
-                DispatchQueue.main.asyncAfter(deadline: .now() + 1) { [weak self] in
-                    guard let self = self else {
-                        return
-                    }
-
-                    self.setupPreviewView(localVideoStream: localVideoStream, retry: retry + 1)
-                }
+                print("Failed to create renderer for lobby video view")
             }
         }
     }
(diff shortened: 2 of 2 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -731,7 +731,14 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     private func promptForTeamsAdmissionDenied() {
         presentAlert(title: "Sorry, you were denied access to the meeting") { [weak self] in
-            self?.navigationController?.popViewController(animated: true)
+            guard let self = self else {
+                return
+            }
+
+            self.navigationController?.popViewController(animated: true)
+            if let lobbyViewController = self.navigationController?.topViewController as? LobbyViewController {
```
AmyL219, line 739: Could this function be moved into LobbyViewController's viewController lifecycle method: ViewWillAppear?

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -75,6 +75,21 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
         return true
     }
 
+    func resetRendererView() {
+        callingContext.withLocalVideoStream { [weak self] localVideoStream in
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1) { [weak self] in
```
AmyL219, line 80: Why do we need to asyncAfter 1 second?

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -99,8 +115,17 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
                 rendererView!.leftAnchor.constraint(equalTo: previewView.leftAnchor).isActive = true
                 rendererView!.rightAnchor.constraint(equalTo: previewView.rightAnchor).isActive = true
                 loadingView.stopAnimating()
+                print("Lobby video view renderer successfully created.")
             } catch {
-                print("Failed to create renderer for lobby video view")
+                print("Failed to create renderer for lobby video view. " + (retry > 0 ? "Retry attempt #\(retry)" : "Attempt to retry..."))
```
jimchou-dev, line 120: Since the denied to join is handled in `promptForTeamsAdmissionDenied` When will this retry be triggered?

cheukchuen: `promptForTeamsAdmissionDenied` leads to `resetRendererView()` being invoked. Within it there's `setupPreviewView` that contains the retry.

AzureCalling/Controllers/CallViewController.swift
cheukchuen: Good point!

AzureCalling/Controllers/LobbyViewController.swift
cheukchuen: It's due to the Calling SDK `VideoStreamRenderer` unavailable for about 0.5 seconds when call ended while video stream is on

AzureCalling/Controllers/LobbyViewController.swift
jimchou-dev: I mean under what condition will it go into the catch block?

cheukchuen: The following statement: `previewRenderer = try VideoStreamRenderer(localVideoStream: localVideoStream)`
`VideoStreamRenderer` is from the Calling SDK

cheukchuen: When the camera is not ready, instantiating `VideoStreamRenderer` throws an error

cheukchuen: To be specific here's the error log:
`Error Domain=ACSException Code=64 "Cannot create multiple renderers for same device or stream."`

jimchou-dev (dismissed):
Tested, LGTM

AzureCalling/Controllers/LobbyViewController.swift
AmyL219: Sorry, But I wonder if it's necessary to set the retrySet to retry the VideoStremRender

I could image the reason why   `previewRenderer = try VideoStreamRenderer(localVideoStream: localVideoStream)` failed is because the local render view inside CallingView is not disposed immediately when when user been kick out from CallingVC. Is that possible that we  manual call local video render dispose function inside `CallingViewController` when it's being dismissed, like the function (ViewWillDisappear)

cheukchuen: Actually that helps! Immediately disposing the renderer for local participant view upon leaving the `CallingViewController` does the job.
Removed the retry mechanism since no longer needed.

AmyL219: Sounds good!

AmyL219 (dismissed):

jimchou-dev (dismissed):

AmyL219 (approved):

jimchou-dev (approved):

PR #55, opened 2021-06-24
Proper behavior when tapping selections and background for bottom drawer

cheukchuen:
## Purpose

* This PR fixes the bug for tapping on participant list item dismisses the list

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Tapping on participant list item does not dismiss the list
* Tapping on participant list background dismisses the list
* Audio device selection remains functional

### DIFF
--- a/AzureCalling/Controllers/BottomDrawerViewController.swift
+++ b/AzureCalling/Controllers/BottomDrawerViewController.swift
@@ -5,20 +5,22 @@
 
 import UIKit
 
-class BottomDrawerViewController: UIViewController {
+class BottomDrawerViewController: UIViewController, UIGestureRecognizerDelegate {
 
     // MARK: Properties
 
     private var tableView: UITableView!
     private var bottomDrawerDataSource: BottomDrawerDataSource?
+    private var allowRowSelection: Bool = false
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
     }
 
-    init(dataSource: BottomDrawerDataSource) {
+    init(dataSource: BottomDrawerDataSource, allowsSelection: Bool) {
         super.init(nibName: nil, bundle: nil)
         self.bottomDrawerDataSource = dataSource
+        self.allowRowSelection = allowsSelection
         self.modalPresentationStyle = .overCurrentContext
     }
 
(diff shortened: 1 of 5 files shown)


AzureCalling/Controllers/BottomDrawerViewController.swift
```diff
@@ -32,24 +35,32 @@ class BottomDrawerViewController: UIViewController {
...
-
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
         openBottomDrawer()
     }
 
+    // MARK: UIGestureRecognizerDelegate events
+
+    @objc func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
```
AmyL219, line 54: Is this necessary? If we don't want to dismiss the viewcontroller when user tap the tableviewcell, could we just set tableview.view.isUserInteractionEnabled to false?

AzureCalling/Controllers/BottomDrawerViewController.swift
```diff
@@ -5,20 +5,23 @@
...
     private var tableView: UITableView!
     private var bottomDrawerDataSource: BottomDrawerDataSource?
+    private var allowRowSelection: Bool = false
 
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
     }
 
-    init(dataSource: BottomDrawerDataSource) {
+    init(dataSource: BottomDrawerDataSource, allowsSelection: Bool) {
```
jimchou-dev, line 20: Can you make the `allowsSelection` variable to have default value `allowsSelection: Bool = false`. So only function calls that needs it enabled need to specify the parameter? (Or vice versa make the default behaviour `allowsSelection: Bool = true`)

AzureCalling/Controllers/BottomDrawerViewController.swift
cheukchuen: Tried `tableView.isUserInteractionEnabled = false`, still triggers `UITapGestureRecognizer` event when tapping the bottom drawer

AzureCalling/Controllers/BottomDrawerViewController.swift
cheukchuen: For sure, default value `false` makes sense since we want the base functionality of a bottom drawer to be just a list

jimchou-dev (approved):
Tested, LGTM

AmyL219 (approved):

PR #63, opened 2021-08-18
Fixing the border still appearing when mute button is turned on

JoshuaLai:
## Purpose
If the mute button is enabled sync the active speak and turn it off. 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
Join call with multiple participants
Start speaking and mute yourself
Notice on the other machine the mute and the active speaker border will sync up 
```

## What to Check
Verify that the following are valid:
Notice on the other machine the mute and the active speaker border will sync up

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -515,8 +515,14 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
             }
 
             participantView.updateDisplayName(displayName: participant.displayName)
-            participantView.updateActiveSpeaker(isSpeaking: participant.isSpeaking)
             participantView.updateMuteIndicator(isMuted: participant.isMuted)
+
+            if participant.isMuted {
+                participantView.updateActiveSpeaker(isSpeaking: false)
+            } else {
+                participantView.updateActiveSpeaker(isSpeaking: participant.isSpeaking)
+            }
+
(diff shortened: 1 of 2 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -515,8 +515,8 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
             }
 
             participantView.updateDisplayName(displayName: participant.displayName)
-            participantView.updateMuteIndicator(isMuted: participant.isMuted)
             participantView.updateActiveSpeaker(isSpeaking: participant.isSpeaking)
```
waynemo, line 518: I think we can move the logic here to call `updateActiveSpeaker(false)` when `participant.isMuted == true`. That way we keep `ParticipantView.updateMuteIndicator` simpler and only concerned with updating the UI according to the argument.

waynemo (changes requested):

AzureCalling/Controllers/CallViewController.swift
JoshuaLai: Updated to no longer have the logic in the participant view

jimchou-dev (approved):
Tested, LGTM

waynemo (approved):

PR #65, opened 2021-10-19
Device icon updates to the appropriate icon when changed

JohnVUgonabo:
## Purpose

* Fixed the device icon to update to the appropriate icon when changed, in both the lobby and calling views

```
[ ] Yes
[ x ] No
```

## Pull Request Type

```
[ x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* Changing the device in the both the lobby and calling view reflects appropriate change to device icon
* When joining call, device selection from lobby is reflected in calling view

## Other Information

Link to bug https://skype.visualstudio.com/SPOOL/_workitems/edit/2508580

### DIFF
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -152,7 +152,10 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     private func openAudioDeviceDrawer() {
         let audioDeviceSelectionDataSource = AudioDeviceSelectionDataSource()
-        audioDeviceSelectionDataSource.didSelectAudioDevice = {
+        audioDeviceSelectionDataSource.didSelectAudioDevice = {[weak self] in
+            guard let self = self else {
+                return
+            }
             self.updateAudioDeviceButtonIcon()
         }
         let bottomDrawerViewController = BottomDrawerViewController(dataSource: audioDeviceSelectionDataSource, allowsSelection: true)
@@ -634,7 +637,8 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         }
     }
     private func updateAudioDeviceButtonIcon() {
-        let deviceIcon = AudioSessionManager.selectedAudioDeviceButtonIcon()
+        let currentAudioDeviceType = AudioSessionManager.getCurrentAudioDeviceType()
+        let deviceIcon = UIImage(named: currentAudioDeviceType.iconName)
         self.verticalSelectAudioDeviceButton.setImage(deviceIcon, for: .normal)
         self.selectAudioDeviceButton.setImage(deviceIcon, for: .normal)
     }
--- a/AzureCalling/Controllers/CallViewController.swift
+++ b/AzureCalling/Controllers/CallViewController.swift
@@ -64,6 +64,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         toggleMuteButton.isSelected = joinCallConfig.isMicrophoneMuted
 
         updateToggleVideoButtonState()
+        updateAudioDevice() // update audio icon at start of call
 
         localParticipantView.setOnSwitchCamera { [weak self] in
             guard let self = self else {
@@ -151,6 +152,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     private func openAudioDeviceDrawer() {
         let audioDeviceSelectionDataSource = AudioDeviceSelectionDataSource()
+        audioDeviceSelectionDataSource.audioDeviceSelectionDelegate = self  //delegate action to AudioDeviceSelectionDataSource
         let bottomDrawerViewController = BottomDrawerViewController(dataSource: audioDeviceSelectionDataSource, allowsSelection: true)
         present(bottomDrawerViewController, animated: false, completion: nil)
     }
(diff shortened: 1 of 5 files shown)


AzureCalling/Controllers/CallViewController.swift
```diff
@@ -629,6 +631,21 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
             bottomDrawerViewController.refreshBottomDrawer()
         }
     }
+    // update the audio device icon with current selection
+    private func updateAudioDevice() {
```
jimchou-dev, line 635: naming too generic it's not actually updating the audio device, should probably include update button icon in the name

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -243,6 +245,19 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
...
+    private func updateAudioDevice() {
+        let currentAudioDeviceType = AudioSessionManager.getCurrentAudioDeviceType() // Get the current audio selection
+        // set icon based on audio type
+        var deviceIcon: UIImage
+        switch currentAudioDeviceType {
+        case .receiver:
+            deviceIcon = UIImage(named: "ic_fluent_speaker_2_28_regular")!
+        case .speaker:
+            deviceIcon = UIImage(named: "ic_fluent_speaker_2_28_filled")!
+        }
```
jimchou-dev, line 258: Similar code with different naming for `deviceIcon` and `image` with CallViewController. updateAudioDevice

Can extract a function to return the UIImage

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -64,6 +64,7 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
         toggleMuteButton.isSelected = joinCallConfig.isMicrophoneMuted
 
         updateToggleVideoButtonState()
+        updateAudioDevice() // update audio icon at start of call
```
jimchou-dev, line 67: Remove the comments (applies for all the files)

AzureCalling/Controllers/datasource/AudioDeviceSelectionDataSource.swift
```diff
@@ -71,3 +72,8 @@ class AudioDeviceSelectionDataSource: NSObject, BottomDrawerDataSource {
         return cell
     }
 }
+
+//this handles updating the device icon when changed
+protocol AudioDeviceSelectionViewControllerDelegate: AnyObject {
+    func updateDeviceAudioSelection() // updates the device audio icon with current Selection
```
jimchou-dev, line 78: Why is this `DeviceAudio` when other places are `audioDevice`. Maybe rename to updateAudioDeviceSelectionButtonIcon or something?

JohnVUgonabo:
@jimchou-dev Thanks for the pointers and corrections, I would make the necessary changes and update this PR

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -124,6 +125,7 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
 
     private func openAudioDeviceDrawer() {
         let audioDeviceSelectionDataSource = AudioDeviceSelectionDataSource()
+        audioDeviceSelectionDataSource.audioDeviceSelectionDelegate = self
```
AmyL219, line 128: I think it's more consistent with the original pattern by using a closure here. 
similar to `dismissDrawer` inside `AudioDeviceSelectionDataSource`, we could have a closure which is defined as 
`var didSelectedItem: (BottomDrawerItem) -> void`  using initializer to define the closure
Inside this closure,  we update the icon on main thread
```
audioDeviceSelectionDataSource.didSelectedItem = {[weak self] item
guard let self = self else {return }
updateAudioDeviceButtonIcon(item.avatar)
}
```

JohnVUgonabo: @AmyL219 Thank you for the feedback and suggestions. Using a closure was a much better approach, trimmed down the amount of codes written. PR now updated

jimchou-dev (dismissed):
Tested, LGTM

AzureCalling/Controllers/CallViewController.swift
```diff
@@ -151,6 +152,9 @@ class CallViewController: UIViewController, UICollectionViewDelegate, UICollecti
 
     private func openAudioDeviceDrawer() {
         let audioDeviceSelectionDataSource = AudioDeviceSelectionDataSource()
+        audioDeviceSelectionDataSource.didSelectAudioDevice = {
+            self.updateAudioDeviceButtonIcon()
```
AmyL219, line 156: Let's use [weak self] here  which prevent retain cycle
audioDeviceSelectionDataSource.didSelectAudioDevice = { [weak self] in 
guard let self = self else {return}
            self.updateAudioDeviceButtonIcon()

AzureCalling/Controllers/LobbyViewController.swift
```diff
@@ -243,6 +247,11 @@ class LobbyViewController: UIViewController, UITextFieldDelegate {
         startCallButton.isEnabled = isDisplayNameValid && isPermissionValid
     }
 
+    private func updateAudioDeviceButtonIcon() {
+        let deviceIcon = AudioSessionManager.selectedAudioDeviceButtonIcon()
```
AmyL219, line 251: same weak self here

AzureCalling/Utilities/AudioSessionManager.swift
```diff
@@ -46,4 +46,15 @@ class AudioSessionManager {
             }
         }
     }
+    public static func selectedAudioDeviceButtonIcon() -> UIImage {
```
AmyL219, line 49: Not sure if it's a good idea to read the icon image from AudioSessionManager. 
Maybe extend the `AudioDeviceType` to be an enum with associated value of iconImage. Rather to put this function here.

JohnVUgonabo: @AmyL219 Thanks for the suggestion. It's a good idea which also helps AudioSessionManager maintain its core and cleans up the code. Suggestions now implemented and PR updated.

AmyL219 (approved):

jimchou-dev (approved):

PR #79, opened 2022-02-08
[Bug] Fix issue with video when users re-enter the Lobby view

vhuseinova-msft:
## Purpose

* Fix the bug that local video automatically on when users re-enter the Lobby view

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
1. Join the Lobby view with Teams link or Meeting ID
2. Start video
3. Tap back button and navigate to the Join Call View
4. Re-enter the Lobby view
```

## What to Check
* The video is turned off
* There is no video preview

### DIFF
--- a/AzureCalling/Controllers/JoinCallViewController.swift
+++ b/AzureCalling/Controllers/JoinCallViewController.swift
@@ -14,7 +14,7 @@ class JoinCallViewController: UIViewController, UITextFieldDelegate {
 
     // MARK: Properties
 
-    var callingContext: CallingContext!
+    var createCallingContextFunction: (() -> CallingContext)!
 
     private var joinCallType: JoinCallType = .groupCall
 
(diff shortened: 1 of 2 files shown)


palatter (approved):
Looks good and tested.

AzureCalling/Controllers/JoinCallViewController.swift
```diff
@@ -101,7 +101,7 @@ class JoinCallViewController: UIViewController, UITextFieldDelegate {
             fatalError("Unexpected destination: \(destination)")
         }
 
-        lobbyViewController.callingContext = callingContext
+        lobbyViewController.callingContext = createCallingContextFunction()
```
AmyL219, line 104: The only question I have to create a new callingContext is if this would ruin the users' `LobbyView` setting when user navigates back from `LobbyView` with inputs `displayName`/`audioSetting`/`speakerSetting` and update the `groupId/team` link in the `JoinView`. When user forward to the `LobbyView` again, the view would lose every previous setting the user input, correct?

vhuseinova-msft: The behaviour for this use case stays the same as it was before the changes: the preselected audio, video and name will be set to default values and the speakerSetting will be set to the preselected user's value (also, this preselected valule is shared across the app).

PR #83, opened 2022-04-14
[BugFix] Add Transport Security Settings in info.plist

AmyL219:
## Purpose

* Fix the issue using `communicationTokenFetchUrl` in `AppSetting.plist` would cause the error: `[The resource could not be loaded because the App Transport Security policy requires the use of a secure connection]`

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


AmyL219:
Just found the error prompting `[The resource could not be loaded because the App Transport Security policy requires the use of a secure connection]` is because my url token is inputted with `http://` by accident. 
Close this PR then

PR #86, opened 2022-07-04
Add AAD login and profile / avatar image retrieval

emlynmac:
## Purpose
Updates the iOS calling hero app calling-ui-library branch with the UI calling library.
This PR is only to cover the log-in cases in this ticket: https://skype.visualstudio.com/SPOOL/_workitems/edit/2912002

```
[x] Yes
[ ] No
```

## Pull Request Type
```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
Get the code, set up your own ACS and AAD services and ensure you can log in. You will also see your avatar picture if you have one, or initials if you don't.

## Other Information
This is an interim PR to a feature branch, created mainly to allow others to share some of the common layout code introduced as we are moving to code-based layout, rather than storyboards.


AzureCalling/External/Authentication/AADAuthHandler.swift
```diff
@@ -53,18 +65,23 @@ class AADAuthHandler {
     }
 
     // MARK: Public API
+    func login(presentVc: UIViewController,
```
emlynmac, line 68: I'm going to simplify the login calls to this one method, so we can get interactive if necessary, otherwise non-interactive.

AzureCalling/Assets.xcassets/Colors/gray900.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
```
Leah-Xia-Microsoft, line 1: We added gray900 here but I don't think we used it anywhere yet.

AzureCalling/Controllers/IntroViewController.swift
```diff
@@ -0,0 +1,262 @@
...
+        let builtWithLabel = FluentUI.Label(style: .body)
+        builtWithLabel.text = "Built with"
+        builtWithLabel.directionalLayoutMargins.top = 24
+
+        let acsImageLabel = FluentUI.Label(style: .button1)
+        acsImageLabel.text = "Azure Communication Services"
+        acsImageLabel.textColor = ThemeColor.primary
+
+        let acsLabelStack = UIStackView(arrangedSubviews: [
+            UIImageView(image: UIImage(named: "acsLogo")),
```
Leah-Xia-Microsoft, line 107: I think we're missing "acsLogo" image in Assets.xcassets which causes the image not being displayed.

AzureCalling/Controllers/IntroViewController.swift
```diff
@@ -0,0 +1,262 @@
+//
```
Leah-Xia-Microsoft, line 1: I don't see that we're locking orientation anywhere but I cannot rotate this screen to landscape mode.

emlynmac: That code is unchanged from the `AppDelegate` file, line 24

AzureCalling/Controllers/IntroViewController.swift
emlynmac: I am contacting Alex about getting it.

AzureCalling/Controllers/IntroViewController.swift
```diff
@@ -0,0 +1,264 @@
...
+        ])
+        userDetails.spacing = 8
+
+        topBar.addSubview(userDetails)
+        userDetails.pinToLeft(withMargin: 16)
+    }
+
+    // MARK: - Authentication State Handling
+    private func handleAuthState() {
+        userAvatar.state.image = authHandler.userAvatar
```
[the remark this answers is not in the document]
emlynmac: Yup, just looking at that now.

AzureCalling/Controllers/IntroViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+        userAvatar.state.primaryText = authHandler.userDisplayName
+        userDisplayName.text = authHandler.userDisplayName
+
+        switch authHandler.authStatus {
```
Leah-Xia-Microsoft, line 171: For the scenario that the user is previously logged in, the authStatus is .unauthorized so the "Sign in" button will appear and after tap on the Sign in button, authHandler.login will find a result and automatically change the view to "Start call" and "Join call". We should somehow check authHandler.login earlier to avoid the extra step of tapping the Sign in button.

AzureCalling/Assets.xcassets/Colors/gray900.colorset/Contents.json
Leah-Xia-Microsoft: Is .primary color using FluentUI's color asset? If so, I believe we don't need to include gray900 in the hero app's asset?

AzureCalling/Assets.xcassets/acsLogo.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
+    {
+      "filename" : "beab6721 a8e3 4493 b90b 351ab8c715db.png",
```
JoshuaLai, line 4: Can we update the file name to align with the image set?

emlynmac: We can but maybe we shouldn't.
This UUID comes from Figma, so probably helps us track down where the design came from. The filename is not referenced in the code.

AzureCalling/Controllers/IntroViewController.swift
emlynmac: Yeah, we could check for an account at the load of the view controller and if so, silently log in.

AzureCalling/Assets.xcassets/Colors/gray900.colorset/Contents.json
emlynmac: Sure, I can remove it for now.

AzureCalling/Assets.xcassets/acsLogo.imageset/Contents.json
emlynmac: Done

Leah-Xia-Microsoft:
Tested with iPhone 11, iOS 15.5

AzureCalling/External/Authentication/AADAuthHandler.swift
```diff
@@ -53,163 +74,173 @@ class AADAuthHandler {
     }
 
     // MARK: Public API
+    func login(presentingVc: UIViewController,
+               completionHandler: @escaping (Error?) -> Void) {
```
JoshuaLai, line 78: Instead of using `Error?` I think we can benefit using `Result`. Success can just be `void` and Failure can provide `Error`

JoshuaLai: Might not even be a bad idea in the Success to return the user information.  That way the AuthHelper doesn't need to save a reference to the user display data (like display name or avatar)

AzureCalling/External/Authentication/AADAuthHandler.swift
```diff
@@ -53,163 +74,173 @@ class AADAuthHandler {
...
             }
+        }
+    }
 
-            guard let result = result else {
-                print("Could not acquire token: No result returned")
-                completionHandler()
-                return
+    func signOut(presentingVc viewController: UIViewController,
+                 completionHandler: @escaping (Error?) -> Void) {
```
JoshuaLai, line 114: Same comment with this closure. Consider using `Result`

AzureCalling/External/Authentication/AADAuthHandler.swift
```diff
@@ -53,163 +74,173 @@ class AADAuthHandler {
...
-            completionHandler()
-
         }
     }
 
-    func signOutCurrentAccount(from viewController: UIViewController,
-                               completionHandler: @escaping () -> Void) {
-        guard let applicationContext = self.applicationContext,
+    // MARK: Private Functions
+    @MainActor
```
JoshuaLai, line 132: Interesting choice marking the method as `MainActor`.  I haven't personally used this attribute, but is there a reason we would want to only have specific methods marked and not the whole class. I imagine all the service calls would require this?

AzureCalling/External/Authentication/AADAuthHandler.swift
```diff
@@ -53,163 +74,173 @@ class AADAuthHandler {
...
 
-    func loadAccountAndSilentlyLogin(from viewController: UIViewController,
-                                     completionHandler: @escaping () -> Void) {
-        loadAccount { [weak self] account in
-            guard let account = account else {
-                completionHandler()
-                return
-            }
+                if result == nil {
+                    result = try await loginInteractively(from: presentingVc)
```
JoshuaLai, line 89: I understand why we would want `loginSilently` to be async. 
But do we want the following code to be async? 
Should we maybe use a dispatchgroup? I'd like to hear your thought process behind this.

AzureCalling/External/Authentication/AADAuthHandler.swift
emlynmac: We could use Result, but as you've pointed out - the return is void, so I don't see the benefit.

AzureCalling/External/Authentication/AADAuthHandler.swift
emlynmac: This _has_ to be this way as the calls to the MSAL library functions that touch the UIKit controls need to be on the main thread. Others do not have this restriction, so there's no need to force the main thread for tasks.

AzureCalling/External/Authentication/AADAuthHandler.swift
emlynmac: Could you expand why you'd want a dispatch group here? 
Using `async` calls makes the code very readable and sequential.

JoshuaLai (approved):

Leah-Xia-Microsoft (approved):
Tested with iPhone 11, iOS 15.5

PR #94, opened 2022-07-21
Migrate the create a call functionality to call the calling library

Leah-Xia-Microsoft:
## Purpose

* Implement a share mechanism for created group ID calls
* Migrate the create a call functionality to call the calling library

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
* Ensure you have set up AAD
* Go to invite screen and observe new design and share functionality

## What to Check
* Start a call


AzureCalling/AppDelegate.swift
```diff
@@ -22,17 +22,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
-
```
AmyL219, line 25: Could you confirm with Alex that if our sample app only supports portrait mode?

AmyL219 (approved):

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -63,537 +30,34 @@ class CallingContext: NSObject {
...
-        })
-    }
-
-    private func setupCalling(completionHandler: @escaping (Result<Void, Error>) -> Void) {
-        callClient = CallClient()
-        self.setupDeviceManager(completionHandler: completionHandler)
-    }
-
-    private func setupCallAgent(displayName: String, completionHandler: @escaping (Result<Void, Error>) -> Void) {
+    func getTokenCredential() -> CommunicationTokenCredential? {
```
JoshuaLai, line 33: @AmyL219 I believe when we were chatting with CallingSDK team they recommended us not pass in `nil` as token. Should we fix this for the hero sample?

AmyL219: Good point. What CallingSDK team said is we should fetch token first. I can schedule a meeeting regarding this

JoshuaLai: Lets not block the PR cause the rest of the changes are good!

JoshuaLai (approved):

AzureCalling/AppDelegate.swift
Leah-Xia-Microsoft: > Could you confirm with Alex that if our sample app only supports portrait mode?

Good point. I'll check with Alex and per discussion, we won't block this PR to wait for a reply. We'll create a seperate PR for supporting landscape if needed.

PR #97, opened 2022-08-02
[Refactor] Improve messaging for missing configuration

emlynmac:
## Purpose
When the token URL endpoint is missing, the app crashes with a forced unwrap of an optional.
This PR adds an assert with details of what needs to be configured.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
Set up with AAD, do not configure an endpoint for the token in the info.plist.

## What to Check
Start a call, observe the app crashing but with an assertion failure and the information message in the console


Leah-Xia-Microsoft (approved):
Thanks for the fix!

PR #103, opened 2022-08-03
Token fetch Refresher optimization

Leah-Xia-Microsoft:
## Purpose

* This refactor is based on the suggestion from the calling SDK team
* for CommunicationTokenRefreshOptions, it's recommended not passing pass initialToken as nil
```
CommunicationTokenRefreshOptions(initialToken: nil, refreshProactively: true, tokenRefresher: tokenRefresher)
```
* The reason being that "the problem is create call agent is blocked unnecessarily if you do it before that you know whether contoso has properly working token refresh url or not instead of trying to debug why createcallagent is stuck"
* Also added activity indicator before launching the call

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## How to Test
* Get the code and set up with AAD.
* Start and Join a group/teams call

## What to Check
*Verify that we can start and join a group/teams call

### DIFF
--- a/AzureCalling/Controllers/InviteViewController.swift
+++ b/AzureCalling/Controllers/InviteViewController.swift
@@ -114,12 +114,11 @@ class InviteViewController: UIViewController {
     }
 
     private func onContinueButtonTapped() {
-        let callConfig = JoinCallConfig(joinId: groupCallId, displayName: displayName ?? "", callType: .groupCall)
-        busyOverlay.presentIn(view: view)
-        self.callingContext.startCallComposite(callConfig, completionHandler: { [weak self] in
-            DispatchQueue.main.async {
-                self?.busyOverlay.hide()
-            }
-        })
+        Task { @MainActor in
+            let callConfig = JoinCallConfig(joinId: groupCallId, displayName: displayName ?? "", callType: .groupCall)
+            busyOverlay.presentIn(view: view)
+            await self.callingContext.startCallComposite(callConfig)
+            self.busyOverlay.hide()
+        }
     }
 }
--- a/AzureCalling/Controllers/InviteViewController.swift
+++ b/AzureCalling/Controllers/InviteViewController.swift
@@ -113,9 +113,8 @@ class InviteViewController: UIViewController {
         self.present(activityVC, animated: true, completion: nil)
     }
 
-    @MainActor
     private func onContinueButtonTapped() {
-        Task {
+        Task { @MainActor in
             let callConfig = JoinCallConfig(joinId: groupCallId, displayName: displayName ?? "", callType: .groupCall)
             busyOverlay.presentIn(view: view)
             await self.callingContext.startCallComposite(callConfig)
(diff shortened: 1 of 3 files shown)


AzureCalling/Controllers/InviteViewController.swift
```diff
@@ -114,6 +115,11 @@ class InviteViewController: UIViewController {
 
     private func onContinueButtonTapped() {
         let callConfig = JoinCallConfig(joinId: groupCallId, displayName: displayName ?? "", callType: .groupCall)
-        callingContext.startCallComposite(callConfig)
+        busyOverlay.presentIn(view: view)
+        self.callingContext.startCallComposite(callConfig, completionHandler: { [weak self] in
+            DispatchQueue.main.async {
```
emlynmac, line 120: You should not need this - onContinueButtonTapped is called from the button action, which is from the main thread.

AzureCalling/Controllers/JoinCallViewController.swift
```diff
@@ -166,7 +167,12 @@ class JoinCallViewController: UIViewController {
             return
         }
         let callConfig = JoinCallConfig(joinId: joinId, displayName: displayName ?? "", callType: joinCallType)
-        callingContext.startCallComposite(callConfig)
+        busyOverlay.presentIn(view: view)
+        self.callingContext.startCallComposite(callConfig, completionHandler: { [weak self] in
+            DispatchQueue.main.async {
```
emlynmac, line 172: Same here, navigateToCall is called from the main thread.
Generally, we don't want to async dispatch unless needed.

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -27,51 +27,71 @@ final class CallingContext {
         self.tokenFetcher = tokenFetcher
     }
 
-    // MARK: Public API
+    // MARK: Private function
+    private func fetchInitialToken(completionHandler: @escaping (String?) -> Void) {
```
emlynmac, line 31: Can we make these `async` methods to follow the rest of the authentication handling?

AzureCalling/External/Calling/CallingContext.swift
```diff
@@ -27,51 +27,71 @@ final class CallingContext {
...
+            } catch {
+                print("ERROR: It was not possible to create user credential.")
+                completionHandler(nil)
+            }
         }
-        return tokenCredential
     }
 
-    func startCallComposite(_ joinConfig: JoinCallConfig) {
+    func startCallComposite(_ joinConfig: JoinCallConfig, completionHandler: @escaping () -> Void) {
```
emlynmac, line 59: This would be greatly simplified to read wth async function.

emlynmac (commented):
I think we can make the functions async to improve readability.
I wonder too, if we should move the silent token refresh down to the AAD handler

Leah-Xia-Microsoft:
> I think we can make the functions async to improve readability. I wonder too, if we should move the silent token refresh down to the AAD handler

I think the tokenFetcher needed for the initialToken fetch is passed down to CallingContext from SceneDelegate. Seems convenient to have the initialToken in CallingContext other than AAD handler?

emlynmac (approved):
LGTM

PR #107, opened 2022-08-05
[Bugfix] Ensure error state is reset and form state is maintained when overridden

emlynmac:
## Purpose
If an override is set from the .xcconfig file, it should keep the state properly in the UI.
This now happens

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Set up the code with a tokenURL
Override the teamsURL in the AppSettings.xcconfig
Go to the join call screen

## What to Check
* Check that the state of the screen shows the correct UI components.
* Switch to group call, observe the clearing of the text and update of controls
*


Leah-Xia-Microsoft (approved):
LGTM

PR #113, opened 2022-08-11
[BugFix] Add link to ACS documentation

emlynmac:
## Purpose
Add a web link to the title on the landing page

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Get the code
* Run the code

## What to Check
* Verify tapping on the "Build with Azure Communication Services" text takes you to the web page

### DIFF
--- a/AzureCalling/Controllers/IntroViewController.swift
+++ b/AzureCalling/Controllers/IntroViewController.swift
@@ -6,7 +6,7 @@
 import FluentUI
 import UIKit
 
-let learnMoreURL = "https://aka.ms/ACS-CallingSample-iOS"
+let learnMoreURL = "https://azure.microsoft.com/services/communication-services/"
 
 class IntroViewController: UIViewController {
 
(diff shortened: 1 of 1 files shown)


AzureCalling/Controllers/IntroViewController.swift
```diff
@@ -6,7 +6,7 @@
 import FluentUI
 import UIKit
 
-let learnMoreURL = "https://aka.ms/ACS-CallingSample-iOS"
+let learnMoreURL = "https://azure.microsoft.com/services/communication-services/"
```
palatter, line 9: Are we switching the from the aka.ms link for a specific reason?

palatter: I've actually never made one before, but if it's possible to update where it points to, it might be better to keep it an aka.ms link. That way we don't need to do a code change to update the link in the future.

emlynmac: I like your thought, but for me the aka.ms tool is in an infinite-redirect-login loop.
We're switching the link away from referencing the code sample (as you've already got that running to see this page in the app).

emlynmac: Alright, got in with edge rather than Safari. Making an aka.ms link

Leah-Xia-Microsoft (approved):

